PR #12, opened 2014-06-12
toArray format and exporting functions

dcousens:
This pull request is intended to be iterative, it currently contains failing tests, and is mostly an inquiry into what is `toArray`'s current behaviour, what it should return, and what should or shouldn't be supported.

The test fixtures are my test fixtures from https://github.com/cryptocoinjs/bigi, and currently contain superfluous data not (yet) necessary for `bn.js`.

@indutny, if you could answer the above question as to what you are happy to support, and what `toArray` is currently meant to be returning (it is not yet tested, nor clear).

The test data currently contains the following expected data:
* `DER` (twos complement big-endian integer representation) data,
* `hex` (unsigned big-endian integer), and
* `hexPadded`: `hex` but with `0` padding out to `64` bytes

The input data:
* `dec` is the signed integer. and
* `udec` is the unsigned equivalent.


dcousens:
I will get back to this soon.

dcousens:
I will get back to this soon and re-evaluate if this is the best way to do it.

PR #30, opened 2015-02-08
Streamlined GCD procedure to be strictly O(N^2) instead of O(N^3)

alexeykudinkin:
2.5x speedup upon 3072b BNs.

PR #31, opened 2015-02-08
Fixed Binary EEA

alexeykudinkin:
`BN._egcd` harness is broken actually. It malfunctions computing inverse of `5` in F(6) as `4`.

P.S. Unfortunately, fix (as expected) slowdowns field inversions by a factor of 2.

### DIFF
--- a/lib/bn.js
+++ b/lib/bn.js
@@ -1488,16 +1488,16 @@ BN.prototype.egcd = function egcd(p) {
     }
   }
 
-  return { 
-    'a': C,
-    'b': D,
-    'gcd': y.ishln(g)
+  return {
+    a: C,
+    b: D,
+    gcd: y.ishln(g)
   };
-}
+};
 
 // This is reduced incarnation of the binary EEA
-// above, designated to invert members of the 
-// _prime_ fields F(p) at a maximal speed 
+// above, designated to invert members of the
+// _prime_ fields F(p) at a maximal speed
 BN.prototype._invmp = function _invmp(p) {
   assert(!p.sign);
   assert(p.cmpn(0) !== 0);
--- a/lib/bn.js
+++ b/lib/bn.js
@@ -1495,6 +1495,55 @@ BN.prototype.egcd = function egcd(p) {
   };
 }
 
+// This is reduced incarnation of the binary EEA
+// above, designated to invert members of the 
+// _prime_ fields F(p) at a maximal speed 
+BN.prototype._invmp = function _invmp(p) {
+  assert(!p.sign);
+  assert(p.cmpn(0) !== 0);
+
+  var a = this;
+  var b = p.clone();
+
+  if (a.sign)
+    a = a.mod(p);
+  else
+    a = a.clone();
+
+  var x1 = new BN(1);
+  var x2 = new BN(0);
+
+  var delta = b.clone();
+
+  while (a.cmpn(1) > 0 && b.cmpn(1) > 0) {
+    while (a.isEven()) {
+      a.ishrn(1);
+      if (x1.isEven())
+        x1.ishrn(1);
+      else
+        x1.iadd(delta).ishrn(1);
+    }
+    while (b.isEven()) {
+      b.ishrn(1);
+      if (x2.isEven())
+        x2.ishrn(1);
+      else
+        x2.iadd(delta).ishrn(1);
+    }
+    if (a.cmp(b) >= 0) {
+      a.isub(b);
+      x1.isub(x2);
+    } else {
+      b.isub(a);
+      x2.isub(x1);
+    }
+  }
+  if (a.cmpn(1) === 0)
+    return x1;
+  else
+    return x2;
+};
+
 BN.prototype.gcd = function gcd(num) {
   if (this.cmpn(0) === 0)
     return num.clone();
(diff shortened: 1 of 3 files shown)


indutny:
I'm afraid `F(6)` is not the field I'm particularly interested in. As far as I am concerned it works quite well on primes. I'd accept PR with `assert(isOdd)`. ;)

Btw, please run `npm test` before pushing it out ;)

alexeykudinkin:
Is `BN._egcd` procedure _provably_ correct for inversions in prime fields?

indutny:
I think it is quite obvious that your algorithm is just an extended version of mine, except last comparison check which decides what to return `x1` or `x2`.

indutny:
The reason I did it this way is that I only wanted it to work on relatively prime numbers, i.e. one `gcd(a, b) === 1`

indutny:
And the last step is a consequence of a different main loop's condition.

alexeykudinkin:
It isn't obvious at all. Compare `if` conditions.

`x = a`, `A = x1`

```
while (a.isEven()) {
   a.ishrn(1);
   if (x1.isEven())
     x1.ishrn(1);
   else
     x1.iadd(delta).ishrn(1);
}
```
```
while (x.isEven()) {
   x.ishrn(1);
   if (A.isEven() && B.isEven()) {
     A.ishrn(1);
     B.ishrn(1);
   } else {
     A.iadd(yp).ishrn(1);
     B.isub(xp).ishrn(1);
   }
}
```

indutny:
Speaking of, I don't think that what you are doing is a right thing. Clearly, if `A.isEven() && B.isOdd()` adding odd number to `A` and shifting it right will loose one bit.

Could you please tell me what exactly is incorrect in my implementation?

alexeykudinkin:
That's not the case: if `A.isEven()` then `B.isEven()` too, or `yp.isEven()`. Algorithm is consistent.

indutny:
Ok, then both algorithms a consistent too. Good news :)

Going to verify my implementation and prove that it is correct. But it appears to be doing a subset of the work that it is done in this PR, doesn't seem like it has any different behavior for relatively prime numbers and odd `b`.

alexeykudinkin:
It's obviously incorrect with even bases. But it's not that easy to say right away whether it is correct for prime bases.

alexeykudinkin:
Yeah its _seems_ to be correct with prime bases. But what i've said may not be seen as a strict prove, rather as a hint why it is so.

alexeykudinkin:
Nope, they do behave directly for _some_ relatively prime numbers like `5` & `6`.

indutny:
What do you mean by `do behave directly`? `5` and `6` are not expected to work fine, since they should be relatively prime and **second one should be odd**.

indutny:
Btw, I think I may rewrite this algorithm in terms of division :) It seems that it'll be faster this way.

alexeykudinkin:
That seems as an obvious contract of `Red.invm` but _not so_ for the `BN.invm`.

alexeykudinkin:
Better not doing so (see https://github.com/indutny/bn.js/pull/30).

indutny:
@alexeykudinkin I see your point. Perhaps we could have `invmp` for `Red.invm` and `invm` for generic case? Do you want to help me with this?

indutny:
> Better not doing so (see #30).

I already did it in 037a551, and got 2x performance. Going to benchmark your PR in a bit.

indutny:
@alexeykudinkin your PR makes `.gcd()` 10x times slower than it is now. I feel like using `.mod()` in `_egcd` will improve it a lot.

alexeykudinkin:
#30? Have benchmarked it yesterday and it's shown up 2.5x boost.

indutny:
@alexeykudinkin hm.... let me paste the numbers.

Here is what it was before 037a551:

```
bn.js#gcd x 220,844 ops/sec ±0.85% (9 runs sampled)
```

Here it is with 037a551:

```
bn.js#gcd x 326,688 ops/sec ±0.48% (9 runs sampled)
```

And here it is with #30:

```
bn.js#gcd x 31,990 ops/sec ±0.90% (9 runs sampled)
```

indutny:
All of this benchmarked on io.js 1.1.0 .

alexeykudinkin:
Can you show the test? How long are numbers in your benchmark?

indutny:
The benchmark is in bd72cce .

indutny:
Ay, my bad. Hold on for a sec.

indutny:
Yeah, benchmark is incorrect. Your implementation is much faster.

indutny:
@alexeykudinkin I just added `.zeroBits()` method. Do you think we could use it to improve `gcd`/`egcd` a bit?

alexeykudinkin:
Not sure if it will make a difference, but let's give it a try.
BTW, Restored `BN._ecgd` as `BN._invmp`. Fixed lint issues.

lib/bn.js
```diff
@@ -1423,7 +1423,82 @@ BN.prototype.divn = function divn(num) {
...
+  var A = new BN(1);
+  var B = new BN(0);
+
+  // C * x + D * y = y
+  var C = new BN(0);
+  var D = new BN(1);
+
+  var g = 0;
+
+  while (x.isEven() && y.isEven()) {
```
indutny: I wonder if we could use `.zeroBits()` here?

indutny: Though, I could do it in a separate PR.

alexeykudinkin: Surely could. I'd rather tried that as a different PR.

indutny:
@alexeykudinkin may I ask you to rebase this? I have moved around the tests and patch does not apply anymore :(

alexeykudinkin:
Done.

indutny:
Landed, thank you!

PR #36, opened 2015-02-12
toArray/toString: add tests

dcousens:
This PR adds tests for `toArray` and `toString` to enforce consistency.
The tests as of right now are failing because I am not aware of the intended output format of `toArray` vs `toString`.

`toString()` seems to output decimal. Tests pass.
`toString(16)` seems to output unsigned hex string with a `-` sign appended to the front.
`toArray()` seems to output unsigned binary, but not in twos complement.

@indutny I'm happy to amend these tests if you could shed some light on what is expected :).


test/fixtures/bn.json
```diff
@@ -0,0 +1,105 @@
...
+    {
+      "dec": "115792089237316195423570985008687907852837564279074904382605163141518161494336",
+      "hex": "fffffffffffffffffffffffffffffffebaaedce6af48a03bbfd25e8cd0364140"
+    },
+    {
+      "dec": "48968302285117906840285529799176770990048954789747953886390402978935544927851",
+      "hex": "6c4313b03f2e7324d75e642f0ab81b734b724e13fec930f309e222470236d66b"
+    }
+  ],
+  "invalid": [
```
dcousens: So far, none of these are throwing.  Is that intentional?

dcousens:
The returned hex from `toString` also seems to be incomplete, in that `hex.length % 2 === 0` doesn't always hold true.  In fact, `0x01` will be returned as `0x1`.  Is it intentional that the hex string is not grouped by bytes?

dcousens:
ping @indutny

test/fixtures/bn.json
```diff
@@ -0,0 +1,105 @@
+{
+  "valid": [
+    {
+      "dec": "0",
+      "hex": ""
```
indutny: wut? Should be `0`, I think.

test/fixtures/bn.json
indutny: Yeah, this is for speed.

test/pummel/dh-group-test.js
```diff
@@ -1,9 +1,10 @@
 var assert = require('assert');
 var BN = require('../../').BN;
-var fixtures = require('../fixtures');
+var fixtures = require('../fixtures/dhGroups');
```
indutny: I'd rather make it `dh-groups`.

test/utils-test.js
```diff
@@ -1,25 +1,39 @@
...
 describe('BN.js/Utils', function() {
   describe('.toString()', function() {
+    fixtures.valid.forEach(function(f) {
+      it('should return ' + f.dec + ' when initialized with ' + f.dec, function() {
+        var bi = new BN(f.dec);
+
+        assert.equal(bi.toString(), f.dec);
+      });
+
+      it('should return big-endian twos complement hex for ' + f.dec, function() {
```
indutny: Maybe test `new BN(f.hex, 16).toString()` too?

test/utils-test.js
```diff
@@ -35,12 +49,41 @@ describe('BN.js/Utils', function() {
...
-        var b = new Buffer(a.toString('hex', 64), 'hex');
+
         assert.equal(a.toString('hex', 64).length, 64);
       });
     });
   });
 
+  describe('.toArray()', function() {
+    fixtures.valid.forEach(function(f) {
+      it('should return a big-endian twos complement integer for ' + f.dec, function() {
```
indutny: 80 column limit?

indutny:
I don't think that we need `invalid` section, at least not now.

indutny:
Otherwise LGTM, thank you! ;)

test/fixtures/bn.json
dcousens: What do you mean, for speed?

test/pummel/dh-group-test.js
dcousens: Sure

test/fixtures/bn.json
indutny: I mean that checking and throwing takes time, and I want it to be as fast as possible.

dcousens: Ok,  what would the expected behaviour be then?  Just so I can put it under these tests :)

dcousens: Like, does it just skip bad characters?

test/utils-test.js
dcousens: Will do

test/fixtures/bn.json
indutny: It is worse than that :) You'll get some garbage. Predictable, but still a garbage.

dcousens: Why not make the default safe, and for the internal usage `BN.createFast` which doesn't do the checks?
Garbage sounds like a real easy way to have people encounter issues without knowing it.  Especially given this library is most often used for crypto...

test/utils-test.js
dcousens: Sure, will fix

test/fixtures/bn.json
indutny: @dcousens I don't even use `.toString()`/`parse` internally, actually. I wonder if the regexp in assert before entering the `_parseHex` and `_parseBase` would be enough? Though, it might be a bit more complicated for `_parseBase`, since the character set may different between different bases.

test/fixtures/bn.json
dcousens: Should it be?  That implies a minimal padding of `1` though right?

test/fixtures/bn.json
dcousens: I think it would most certainly be a start.
I'd have to look at the implementations for other ways to maintain performance and increase safety.

dcousens:
Ok @indutny I addressed most of the nits. 

Waiting on your answer about whether it should have a the default zero padding of `1` (which IMHO I think is odd).

Also curious if after the discussion about regex's whether I keep/remove the `invalid` section.

test/fixtures/bn.json
indutny: I'd like to choose the path of least surprise. bn.js is used not only for crypto purposes, but for UI input/output too.

dcousens: Granted, but for UI output you would surely have a padding of `1` then.  I don't see why it should be a default otherwise?

dcousens: I'm not sure on this one,  I may be biased by my work with previous JS libraries.
Do we have any others with which to measure against?

dcousens: From a quick look around, it seems the default is to print the `0`.  

https://github.com/node-modules/node-biginteger/
https://github.com/peterolson/BigInteger.js

I'm still not sure what I agree with, but I'll happily change the test based on a semi-consistent consensus.

dcousens: Fixed the test @indutny, after some further research and thinking about it.  I'm happy with a consistent padding of at least `1`.  
Conceptually I could understand an empty string for zero,  as the absence of any information, but, in this case it wasn't consistent with the decimal result so whatever.

test/fixtures/bn.json
```diff
@@ -0,0 +1,105 @@
...
+      "dec": "0",
+      "hex": "00"
+    },
+    {
+      "dec": "1",
+      "hex": "01"
+    },
+    {
+      "dec": "-1",
+      "hex": "ff"
```
indutny: Why should this happen? Hex could be negative too, right?

indutny:
@dcousens looking very good now, except positive-only hex and a failing tests... Speaking of this, I'm not a big fan of having failing tests in repo's master. Would you mind separating them into an another PR, where we'll figure out the best way to make it work properly?

test/fixtures/bn.json
dcousens: It is the twos complement, that is negative.

dcousens:
Sure, did you have a branch I could target?

test/fixtures/bn.json
indutny: Mmm...? I don't think I support twos complement anywhere in code. Also, it doesn't seem to be a desired feature to me.

indutny:
@dcousens oh right! Added you as a collaborator to this repo, I hope you don't mind :)

test/fixtures/bn.json
dcousens: I guess it depends how you want this to be used.
You block out the ease of use as a data interchange format if you use '-'
over a sign bit (twos complement).

Otherwise I have to either encode it myself,  or just send the string
itself.
Using twos complement is the standard for this,  as that is it's entire
purpose.

Also,  you already would be encoding twos complement for use in elliptic?
(DER encoding uses twos complement integers)

indutny: @dcousens HEX is nothing special to `.toString()`, the base-36 or base-3 encoding should work the same way as the HEX does.

I guess the two's complement thing should happen in the code responsible for encoding this. There are various way to encode it, and this doesn't look like a reason to make HEX something special.

dcousens: Agreed,  maybe it's worth having it as separate functionality?  In any
case,  I'll amend the test vectors.

indutny: Sure, makes sense for another issue/PR ;)

axic:
@dcousens do you think it would make sense revisiting this PR?

jprichardson:
@axic I sure hope so.

> Garbage sounds like a real easy way to have people encounter issues without knowing it. Especially given this library is most often used for crypto.
I think it is likely people will just accept the garbage result as what it should be at face value, because the underlying implementation they have no understanding of.

Couldn't agree more @dcousens!

axic:
Some of the things you've mentioned could be addressed/discussed as part of #129:

> toString(16) seems to output unsigned hex string with a - sign appended to the front.
> toArray() seems to output unsigned binary, but not in twos complement (the standard)

dcousens:
Added this back into my work queue :

dcousens:
re-adding to work queue attempt 3

earonesty:
Note: right now the toString() function doesn't zero pad hex (or anything > base 10), which results in strings like "a" instead of "0a" being output as valid hex strings.   Perhaps this is causing tests to fail.   If people are relying on the current behavior, might need to make "padding" an option.

DONALDSULLENDER (approved):

DONALDSULLENDER (approved):

PR #43, opened 2015-03-12
Update bn.js

Yaffle:


dcousens:
@Yaffle for those of us who aren't fully aware of the context, what does this PR do?

Yaffle:
In short: It does nothing.

`_parseBase` and `toString` use different names for variables with same meaning: `groupLength` and `limbLen` and `groupSize` and `limbPow`. I am changing that to use same names. Also, both methods could use same algorithm to compute values for those variables.

indutny:
@Yaffle may I ask you to write proper commit log? `Update bn.js` is kind of vague

PR #46, opened 2015-04-29
Input for numbers >=  2 ^ 52 and < 2 ^ 53 -1

braydonf:


indutny:
bn.js does not accept floating point input, sorry!

braydonf:
The equivalent as a string works fine:
```js
var num = (50000000).toString(10);
var b = new BN(num, 10);
assert.equal(b.toString(10), num);
```

indutny:
Well, it is not the same. The string has all decimal places filled, while the floating point number is just a representation.

indutny:
See, the purpose of this library is to have a integer precision, so it works with and accept only integers.

braydonf:
Should the input for BN be required to be a string then?

indutny:
On one hand - yes, on other hand it is still slower than passing 32bit integers directly. I'd rather leave it as it is for performance reasons.

indutny:
Perhaps, this thing should be documented... But we don't have docs yet :)

braydonf:
This also works:
```js
var num = 45000000 * 1e8;
assert.equal(new BN(num, 10).toString(10), num.toString(10));
```

gabegattis:
I made a script to find the failing number.

4503599627370495 works as expected
4503599627370496 fails

```
var badNumber1 = 4503599627370496;
var bn1 = new BN(badNumber1, 10);
bn1.toString(10); // this is '0'

var badNumber2 = 4503599627370497;
var bn2 = new BN(badNumber2, 10);
bn2.toString(10); // this is '1'

var badNumber3 = 4503599627370498;
var bn3 = new BN(badNumber3, 10);
bn3.toString(10); // this is '2'

var badNumber3 = 4503599627370499;
var bn3 = new BN(badNumber3, 10);
bn3.toString(10); // this is '3'
```

gabegattis:
4503599627370496 is 16^13

indutny:
Yep, this is the maximum precision that you may have with the integers in javascript ;)

braydonf:
```js
assert.equal(Number.MAX_VALUE > Math.pow(2, 52), true)
```

indutny:
@braydonf Number.MAX_VALUE is a maximum floating point value. The number of precision bits is 52bit.

gabegattis:
https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Number/MAX_SAFE_INTEGER

max integer is 2^53 - 1

indutny:
Mmmm... this is what I am talking about, right?

gabegattis:
(2^53) - 1 = 9007199254740991
2^52        = 4503599627370496

indutny:
@gabegattis I said that it has 52 bits of precisions, this is exactly how much (2^53) - 1 has.

gabegattis:
Right, but this feature in BN only seems to support 51 bits.

The highest number that worked was (2^52) - 1.
The next number, 2^52, did not work as expected.

lib/bn.js
```diff
@@ -61,12 +62,19 @@ BN.prototype._init = function init(number, base, endian) {
...
       this.words = [
         number & 0x3ffffff,
         (number / 0x4000000) & 0x3ffffff
       ];
       this.length = 2;
+    } else {
+      this.words = [
+        number & 0x3ffffff,
+        (number / 0x4000000) & 0x3ffffff,
+        (number / 0x10000000000000) & 0x3ffffff
```
indutny: I wonder if we could be more explicit in a mask here? It is just `1`, right?

indutny:
One minor comment, otherwise LGTM. Thank you very much for finding this and convincing me that this should be fixed ;)

lib/bn.js
braydonf: Yep, it's `1`

dcousens:
Looks good to me,  meant to bring this up myself a while ago.

indutny:
Landed in 64d710d7babeb55f5e491976964653751282e23c, with a very minor fix. Thank you everyone!

Issue #47, opened 2015-06-06
-178 = 10 * (-17) + 2, -178 = -10 * 17 + 8


Yaffle:
Your `mod` result is always positive and division is a truncated division, so the quotient `q` and the remainder `r` of `a` divided by `n` do not satisfy `a = nq + r`.
see http://en.wikipedia.org/wiki/Modulo_operation#Remainder_calculation_for_the_modulo_operation

I think, https://github.com/justmoon/node-bignum is correct here.

dcousens:
@Yaffle care to add example code?

Yaffle:
@dcousens ,

```
      var a = "-178";
      var n = "10";
      var q = new BN(a, 10).div(new BN(n, 10)).toString();
      var r = new BN(a, 10).mod(new BN(n, 10)).toString();
      console.log(`${a} = ${n} * ${q} + ${r}`);
```

indutny:
Should be fixed by https://github.com/indutny/bn.js/pull/54 . PTAL.

@dcousens I need to consult with you on this a bit. It is a breaking change, and I wonder if there might be any modules that will need to be updated to support this one. The main difference is that `.mod()` will now return negative number, when the input was negative itself.

There was few places in `elliptic` that needs to be fixed, but it wasn't that complicated (considering that I have pretty big test case).

What are your thoughts? Is it worth fixing, at the price of potentially breaking user code in un-obvious places?

dcousens:
This would break a lot of existing ECDSA code where we are relying on this behaviour currently.
Several of my use cases are relying on this behaviour (though, they are using a different BN library right now),  so I'd agree this could have a potentially disastrous effect on users,  at best its absolutely a major version bump IMHO.

@calvinmetcalf thoughts,  I'm pretty sure we're relying on this in crypto-browserify throughout?

indutny:
@dcousens I have a pending fix for elliptic library, but this will require major bump on it as well. I guess we can deal with it.

Still would like to hear @calvinmetcalf thoughts on this.

jprichardson:
Doesn't it make sense to make the fix for the sake of correctness and predicability and to just bump the major? Maintaining a changelog would help to communicate these sorts of issues for anyone that wants to upgrade major versions.

indutny:
@jprichardson the problem is that checking it might be not trivial. But I agree that it should be done.

Yaffle:
<table>
<tr>
<th>a</th>
<th>n</th>
[...]

PR #54, opened 2015-07-04
lib: fix divmod signs

indutny:
Return negative `mod` to make sure that:

    a = Q * b + R

Fix: #47

### DIFF
--- a/lib/bn.js
+++ b/lib/bn.js
@@ -559,9 +559,7 @@ BN.prototype.neg = function neg() {
 
 
 // Or `num` with `this` in-place
-BN.prototype.ior = function ior(num) {
-  this.sign = this.sign || num.sign;
-
+BN.prototype.iuor = function iuor(num) {
   while (this.length < num.length)
     this.words[this.length++] = 0;
 
@@ -569,6 +567,11 @@ BN.prototype.ior = function ior(num) {
     this.words[i] = this.words[i] | num.words[i];
 
   return this.strip();
+};
+
+BN.prototype.ior = function ior(num) {
+  assert(!this.sign && !num.sign);
+  return this.iuor(num);
 };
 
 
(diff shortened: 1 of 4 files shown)


Yaffle:
this branch passes my tests

test/red-test.js
```diff
@@ -72,10 +72,23 @@ describe('BN.js/Reduction context', function() {
         var p = new BN(257);
         var m = fn(p);
         var a = new BN(3).toRed(m);
-        var b = a.redInvm(p);
+        var b = a.redInvm();
         assert.equal(a.redMul(b).fromRed().toString(16), '1');
       });
 
+      it('should invm numbers (regression', function() {
```
dcousens: Missing )

indutny:
@dcousens fixed

indutny:
@dcousens how about this? Should fix all mentioned issues (by throwing :D )

dcousens:
@indutny hahaha.
That works,  just are you sure you want to expose the unsigned methods?  That means supporting them later.

indutny:
@dcousens I'm going to use them in elliptic. They appear to be very useful in the reduction-only case. Supporting them does not seem to be a big problem, considering that signed versions are most likely going to fall back to the unsigned ones in some (if not all) cases.

dcousens:
Solid,  LGTM :+1:

lib/bn.js
```diff
@@ -609,9 +626,16 @@ BN.prototype.and = function and(num) {
...
+    return this.clone().iuand(num);
+  else
+    return num.clone().iuand(this);
+};
+
 
 // Xor `num` with `this` in-place
-BN.prototype.ixor = function ixor(num) {
+BN.prototype.iuxor = function iuxor(num) {
   this.sign = this.sign || num.sign;
```
dcousens: Given that this is `unsigned`,  does this logic still make sense?

indutny: It doesn't, thanks.

indutny: Fixed.

indutny:
Landed and released! Thank you!

ashnur:
This bit me.

indutny:
@ashnur how come? It is fixed now.

ashnur:
This was written for 2.0.5 https://github.com/rat-nest/big-rat/blob/master/lib/num-to-bn.js#L13 and when I updated to latest it broke. I haven't written the original, so I am not sure if it was used correctly. I rewrote the code to use `.ushln` instead which passes the tests, but it will take me some time to be sure if everything is correct.

indutny:
Aaah, I see now. I'm terribly sorry for this, but it has to be done this way.

PR #55, opened 2015-07-16
Add `padTo` option to toArray

sublimator:


indutny:
Landed and released, thanks!

indutny:
@sublimator may I ask you to create a PR for readme as well? Sorry forgot to say this before landing :(

sublimator:
How do you feel about allowing the padTo to work as the first arg, while I'm at it?

indutny:
@sublimator this will break stuff :( Let's keep it as it is

PR #56, opened 2015-07-16
readme: update .toArray

sublimator:


sublimator:
Something to consider: currently this does nothing about when the array is already larger than padTo.

indutny:
Hell yes, thanks!

Interesting, maybe we should make it throw? I doubt that the code that works with this will ever handle such results properly.

indutny:
Landed in 495165a

sublimator:
Throwing sounds good to me

sublimator:
Opened a pull request with strict length semantics (renamed padTo to length in the process)

PR #59, opened 2015-08-25
bn: add base 2 padding for .toString()

braydonf:


indutny:
Thank you, landed and released!

PR #61, opened 2015-09-17
lib: .pseudoRandom(min, max)

Ianis:
Add the possibility to generate big random numbers.

### DIFF
--- a/lib/bn.js
+++ b/lib/bn.js
@@ -2403,27 +2403,24 @@ Mont.prototype.invm = function invm(a) {
 };
 
 //
-// Random number generator
+// Pseudo random number generator
 //
 
-BN.random = function random(min, max) {
-  var range = max.sub(min);
+BN.pseudoRandom = function pseudoRandom(min, max) {
+  var n = max.bitLength();
+  var len = Math.floor((n + 64 - 1) / 0xf) + 1;
   var out = new BN(null);
-  out.length = range.length;
-  out.words = [];
+  out.words = new Array(len);
+  out.length = len;
 
-  for (var i = range.length - 1, restricted = true; i >= 0; i--) {
-    var top = restricted ? range.words[i] : 1e7;
-    var word = Math.round(Math.random() * top);
-    out.words.unshift(word);
-    if (word < top)
-      restricted = false;
-  }
+  for (var i = 0; i < len; i++)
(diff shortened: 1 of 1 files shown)


lib/bn.js
```diff
@@ -2402,4 +2402,28 @@ Mont.prototype.invm = function invm(a) {
...
+//
+
+BN.random = function random(min, max) {
+  var range = max.sub(min);
+  var out = new BN(null);
+  out.length = range.length;
+  out.words = [];
+
+  for (var i = range.length - 1, restricted = true; i >= 0; i--) {
+    var top = restricted ? range.words[i] : 1e7;
```
indutny: What is `1e7` here?

lib/bn.js
```diff
@@ -2402,4 +2402,28 @@ Mont.prototype.invm = function invm(a) {
...
 
+//
+// Random number generator
+//
+
+BN.random = function random(min, max) {
+  var range = max.sub(min);
+  var out = new BN(null);
+  out.length = range.length;
+  out.words = [];
```
indutny: Let's create an array and fill it. Unshifting is very slow.

Ianis:
It's actually an old code and I just noticed big mistakes. 
The `1e7 ` was supposed to fix it but it make it worse in some specific conditions.

dcousens:
I think you should be very careful about this,  given this libraries predominant usage in cryptography applications.

Perhaps `insecureRandom`?

indutny:
@dcousens pseudoRandom?

dcousens:
@indutny sure, but it is a very poor `pseudoRandom`.

Ianis:
@dcousens Yes indeed Furthermore JavaScript's `Math.random()` function is not considered a cryptographically-secure random number generator.

indutny:
@dcousens makes sense... do you think it might be a better idea to release it as a separate module instead?

dcousens:
@indutny if it is not to be inlined, then you can easily just use the `randombytes` module that already exists and just import it as hex.

Ianis:
@dcousens  You mean it should support `crypto.randomBytes()` and `window.crypto.getRandomValues()` depending on whether it is running in node or in a browser ?

indutny:
I think it should go into separate module...

indutny:
Sorry!

PR #62, opened 2015-10-27
bn: `.sign` either `0` or `1`, other improvements

indutny:
Make `.sign` a plain integer instead of boolean. Possible values are
either `0` or `1`.

Other improvements:

- Peel first iteration of `smallMulTo`. This gives about 5% boost to
  multiplication
- Cast array loads to small integers, this gives about 5% boost to
  addition and probably other routines too

cc @calvinmetcalf This is probably going to be a major version bump. What are your thoughts on this?

### DIFF
--- a/lib/bn.js
+++ b/lib/bn.js
@@ -29,7 +29,7 @@ function BN(number, base, endian) {
     return number;
   }
 
-  this.sign = false;
+  this.sign = 0;
   this.words = null;
   this.length = 0;
 
@@ -87,7 +87,7 @@ BN.prototype._init = function init(number, base, endian) {
     this._parseBase(number, base, start);
 
   if (number[0] === '-')
-    this.sign = true;
+    this.sign = 1;
 
   this.strip();
 
@@ -99,7 +99,7 @@ BN.prototype._init = function init(number, base, endian) {
 
 BN.prototype._initNumber = function _initNumber(number, base, endian) {
   if (number < 0) {
-    this.sign = true;
+    this.sign = 1;
     number = -number;
   }
   if (number < 0x4000000) {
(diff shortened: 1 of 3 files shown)


calvinmetcalf:
either do it soon so we can fold it into the current updates or take your time and land indutny/elliptic#58 first so we can let the dust settle a bit in crypto-browserify land

dcousens:
ACK

indutny:
@dcousens what do you think about the changes? +1, -1, or 0?

dcousens:
I think the term `sign` is the only problem here,  its ambiguous as the sign can represent both positive and negative connotations.

`.negative` is what the variable is actually representing.

lib/bn.js
```diff
@@ -812,17 +816,17 @@ BN.prototype.add = function add(num) {
...
+    num.sign = 0;
     var r = this.iadd(num);
-    num.sign = true;
+    num.sign = 1;
     return r._normSign();
 
   // -this - num = -(this + num)
-  } else if (this.sign) {
-    this.sign = false;
+  } else if (this.sign !== 0) {
```
dcousens: Any reason to use `!== 0` vs `=== 1`?   You alternate between the two?

indutny: Mistake, fixed.

indutny:
@dcousens actually it makes total sense, renamed `sign` to `negative`.

indutny:
@calvinmetcalf LGTY?

dcousens:
@indutny I'm not seeing the new changes?

indutny:
@dcousens ooops, git prompted password and I didn't see it. Here you go

dcousens:
ACK, LGTM.

indutny:
Landed and released major version.

Issue #63, opened 2015-10-29
regression in 4.1.0


calvinmetcalf:
specifically in dsa calculations, will try to make a test case, but basically browserify-sign dsa doesn't work with 4.1.0 but does with 4.0.0

calvinmetcalf:
@indutny where does [this 26](https://github.com/indutny/bn.js/blob/1dd4cfe431392dac557e98c1d0e8610bb710092b/lib/bn.js#L2369) come from? will it possibly not work for big enough numbers?

indutny:
@calvinmetcalf it is number of bits in a single limb. This is the same for all numbers.

calvinmetcalf:
hm I am reliably getting a failure in sign-browserify but with the same `this.m`, a, and num get a different output from red.prototype.pow then when tested directly is there another variable i'm missing which is affecting the output?

calvinmetcalf:
fixed

PR #64, opened 2015-10-29
failing test for the 4.1.0 reversion

calvinmetcalf:
basically mont and red are not getting the same results


calvinmetcalf:
re  #63

indutny:
argh... thanks for finding this out

indutny:
It doesn't look like it was working before, either... can you confirm?

indutny:
Wait, of course they are different. You are comparing representations in reduction context. Try calling `fromRed()` on them ;)

calvinmetcalf:
damn you are right back to square one

PR #65, opened 2015-10-31
`imul`, `isqr` fixes

alexeykudinkin:
First, `isqr` isn't in-place in current implementation.
Making it in-place, however reveals that `imul` implementation (which in itself looks like a clone of `bigMulTo`) is flawed: therefore its replacement with single entry-point seems reasonable enough for me.

Cons?

### DIFF
--- a/lib/bn.js
+++ b/lib/bn.js
@@ -1023,51 +1023,7 @@ BN.prototype.mul = function mul(num) {
 
 // In-place Multiplication
 BN.prototype.imul = function imul(num) {
-  if (this.cmpn(0) === 0 || num.cmpn(0) === 0) {
-    this.words[0] = 0;
-    this.length = 1;
-    return this;
-  }
-
-  var tlen = this.length;
-  var nlen = num.length;
-
-  this.negative = num.negative ^ this.negative;
-  this.length = this.length + num.length;
-  this.words[this.length - 1] = 0;
-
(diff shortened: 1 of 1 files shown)


alexeykudinkin:
Holy grail, wrong base! Sorry for that.

alexeykudinkin:
That way looks much better.

lib/bn.js
```diff
@@ -1104,7 +1060,7 @@ BN.prototype.sqr = function sqr() {
 
 // `this` * `this` in-place
 BN.prototype.isqr = function isqr() {
-  return this.mul(this);
+  return this.imul(this.clone());
```
indutny: I agree with this, but disagree with above.

indutny:
@alexeykudinkin is there any kind of regression test or anything that demonstrates `imul` problems? I'd rather fix current `imul` implementation.

alexeykudinkin:
For what purpose there're separate implementations for `imul` & `bigMulTo`?

Regression: Patch `isqr` the way described & run `npm test`.

indutny:
It is for performance reasons, it should not allocate anything.

`smallMulTo` is optimized for numbers with low amount of limbs, and is not correct for bigger numbers.

indutny:
Argh, I know what's happening. It needs a `bigMulTo`. Let's ditch it. LGTM

indutny:
Landed in b627b19, thank you!

PR #67, opened 2015-11-02
lib: 20% speed-up for `mul` on 256-bit numbers

indutny:
Unroll loop, inline all operations, use small integers. Boosts
performance by 20%.

cc @dcousens @calvinmetcalf 

I have a problem with this PR that I would like to discuss. It improves the performance at the cost of the bloated code base. Even after the uglify-js compression the code is ~59% bigger than the code before it (compressed too).

I think this should go into a separate module, but we don't have infrastructure for this. What do you think will be best:

* Separate npm module
* Separate files and `dist/` folder
* Something else?

Thanks!


calvinmetcalf:
What are the absolute sizes this library isn't huge so what does 59% mean?

indutny:
@calvinmetcalf  46040 bytes vs 28449 bytes. Quite big increase

calvinmetcalf:
- so gzipped it's 7717 bytes vs 10419 so 35% increase
- Math.imul is not universally supported so that would be a thing against
- if this was to switch over to using browserify there would be some opportunities for size reduction if that is an issue
- I'm frankly surprised gzip doesn't do a better job on it considering how it's pretty much the same thing several times
- would you get some sort of similar speedup if it was a pure function call that v8 could inline easily if this heats up?

calvinmetcalf:
wouldn't the Function constructor make it ever faster/smaller

indutny:
It is optional depending on the presence of Math.imul.

Function can't be used in heavily-sandboxed environments, like Chrome's App
Store.

dcousens:
Are you sure the performance increase is from the unrolling solely, and not just from the hard-coded assumptions? (256-bit)
I think you could get similar performance without the code bloat,  but I don't have time to play with it right now.

indutny:
@dcousens the performance comes from eliminating bounds checks, ordering loads properly, and performing most of the math on 13-bit integers.

dcousens:
@indutny awesome work providing the code gen too.

calvinmetcalf:
the elimination of bounds checks, the load ordering and the 13 bit math could likely be done in a series of function calls that would be MUCH easier to maintain in addition to being smaller

indutny:
@calvinmetcalf nope they can't, unless you know how ;)

dcousens:
@indutny across what engines are we benching?  Are we just using https://github.com/indutny/bn.js/blob/master/benchmarks/index.js ?

indutny:
I'm using this with elliptic:

```js
var BN = require('bn.js');
var elliptic = require('./');

var ecdsa = new elliptic.ec('secp256k1');

var keys = ecdsa.keyFromPrivate('b780239d832dfa8e09a3badfc7ee2f3557d5c0320ca2a4db22a16b946e481170', 'hex');

var msg = new BN('b780239d832dfa8e09a3badfc7ee2f3557d5c0320ca2a4db22a16b946e481170', 16);
var signature = ecdsa.sign(msg, keys);
console.log(ecdsa.verify(msg, signature, keys));

// Warm-up
for (var i = 0; i < 100; i++) {
  ecdsa.verify(msg, signature, keys);
  ecdsa.sign(msg, keys);
}


console.time('verify');
for (var i = 0; i < 3000; i++)
  ecdsa.verify(msg, signature, keys);
console.timeEnd('verify');

console.time('sign');
for (var i = 0; i < 3000; i++)
  ecdsa.sign(msg, keys);
console.timeEnd('sign');
```

indutny:
Testing only on V8 so far.

dcousens:
Happy to play with this, but won't be able to for a few days.
Do you have a time constraint on when you want this merged @indutny?

indutny:
@dcousens no constraints! ;)

indutny:
Going to land it as it is. Let's figure out size issues later.

indutny:
Landed in 0b05b44, thank you!

PR #68, opened 2015-11-03
rough idea how to make it less verbose

calvinmetcalf:
haven't checked perf


indutny:
This is not going to work. Here are results with your patch and with master (from elliptic):

```
# with patch
verify: 10545.672ms
sign: 3968.732ms

# without
true
verify: 6560.543ms
sign: 2792.045ms
```

indutny:
The reasons are, accessing context variable `c`. Placing function inside of that function, so it can't be inlined efficiently.

calvinmetcalf:
ok rejiggered it so it was inlineable but it hits the 'cumulative AST node limit'

dcousens:
> 'cumulative AST node limit'

How are you measuring this OOI?

calvinmetcalf:
was running it in node with the --trace-inline flag

dcousens:
Closing unless someone else wants to take up the reigns

PR #71, opened 2015-11-21
bn: introduce .isNeg()

axic:


indutny:
Looks like the lint is no longer passing after your latest commit.

indutny:
Otherwise, LGTM. Please let me know once it will pass again, and I will take another look. Thanks!

axic:
Honestly I don't think that is caused by my change. ```npm test``` runs without failures on my setup (which uses node v0.12.7). For the sake of testing I've also installed v0.10.40 and that works too.

indutny:

axic:
Thank you!

PR #74, opened 2015-11-22
Introduce .toNumber()

axic:
A simple wrapper around toString() to get the an actual number object in JS. Throws an error if its not safe to represent in JS.

### DIFF
--- a/lib/bn.js
+++ b/lib/bn.js
@@ -452,7 +452,7 @@ BN.prototype.toString = function toString(base, padding) {
 
 BN.prototype.toNumber = function toNumber() {
   assert(this.bitLength() <= 53, 'Number can only safely store up to 53 bits');
-  return parseInt(this.toString());
+  return parseInt(this.toString(), 10);
 };
 
 BN.prototype.toJSON = function toJSON() {
--- a/lib/bn.js
+++ b/lib/bn.js
@@ -450,6 +450,11 @@ BN.prototype.toString = function toString(base, padding) {
   }
 };
 
+BN.prototype.toNumber = function toNumber() {
+  assert(this.bitLength() <= 53, 'Number can only safely store up to 53 bits');
+  return parseInt(this.toString());
+};
+
 BN.prototype.toJSON = function toJSON() {
   return this.toString(16);
 };
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,6 +1,15 @@
+sudo: false
 language: node_js
 node_js:
-  - "4.1"
-branches:
-  only:
-    - master
+  - "0.10"
+  - "0.12"
+  - "4"
+  - "5"
+env:
+  matrix:
+    - TEST_SUITE=test
+matrix:
+  include:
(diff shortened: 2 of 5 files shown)


lib/bn.js
```diff
@@ -450,6 +450,11 @@ BN.prototype.toString = function toString(base, padding) {
   }
 };
 
+BN.prototype.toNumber = function toNumber() {
+  assert(this.bitLength() <= 53, 'Number can only safely store up to 53 bits');
+  return parseInt(this.toString());
```
indutny: `parseInt(..., 10)`?

indutny:
And use case is?

lib/bn.js
axic: That will allow bitLength()>53 numbers to be wrongly transformed into number. JS won't complain, see http://stackoverflow.com/questions/307179/what-is-javascripts-highest-integer-value-that-a-number-can-go-to-without-losin

axic:
It is a shorthand to parseInt + BN.toString with overflow checks.

I'm using it in a library for handling Ethereum's ABI.

lib/bn.js
indutny: I wasn't suggesting to remove `assert`

indutny:
I'm not sure if it is really useful for everyone... cc @dcousens

indutny:
cc @calvinmetcalf

lib/bn.js
axic: Ah, I've assumed parseInt is radix 10 by default, but it is radix 8 if it starts with '0'?

indutny: Exactly.

axic: toString() defaults to radix 10 and padding 1. In what case would it result in leading zeroes (apart from the number 0)?

indutny: Nope, it won't result in leading `0`

calvinmetcalf:
I think this could be useful, for extracting a number and avoiding accidentally getting a result which is not as precise as you think.

indutny:
This patch does not apply cleanly to master. May I ask you to run:

```
git fetch origin
git rebase origin/master
```

And resolve all conflicts?

dcousens:
Couldn't you just use `parseInt`?
Could that be less accurate than expected?

axic:
@dcousens it was changed to parseInt(.., 10) on the request of @indutny assuming that in some hypothetical cases toString() could return with a leading zero and an ancient Javascript interpreter might consider that as octal. Probably safer this way.

axic:
@indutny did the rebase

indutny:
It was still messed up, but I figured it out ;)

Thanks, landed!

axic:
I need to clean up my tree as the previous pull request was cherry picked and that seems to have messed it up.

indutny:
Yeah, I always cherry-pick.

axic:
Thank you for the merge! I'll just recreate my fork to avoid any future mess as there's no other pull request pending.

PR #75, opened 2015-11-22
travis: use 0.10/0.12/4/5

fanatid:


indutny:
Landed in 3dc8f14, thank you!

PR #76, opened 2015-12-01
Add .toBuffer() to return a node.js Buffer

axic:
As the constructor already supports Buffer as an input, it would make sense to make this bidirectional.

Buffer+BN together is used extensively in ethereumjs-vm and ethereumjs-abi to encode/decode bytecode.

Added the .toBuffer() as a new feature which asserts on Buffer being present. Comes with its own test.

Also added tests for using Buffer as an input to the constructor.

### DIFF
--- a/lib/bn.js
+++ b/lib/bn.js
@@ -53,7 +53,7 @@
 
   var Buffer;
   try {
-    Buffer = require('buf' + 'fer').Buffer;
+    Buffer = require('buffer').Buffer;
   } catch (e) {
   }
 
--- a/lib/bn.js
+++ b/lib/bn.js
@@ -50,7 +50,12 @@
 
   BN.BN = BN;
   BN.wordSize = 26;
-  BN._hasBuffer = typeof Buffer !== 'undefined';
+
+  try {
+    BN._Buffer = require('buf' + 'fer').Buffer;
+  } catch (e) {
+    BN._Buffer = undefined;
+  }
 
   BN.max = function max (left, right) {
     if (left.cmp(right) > 0) return left;
@@ -497,8 +502,8 @@
   };
 
   BN.prototype.toBuffer = function toBuffer (endian, length) {
-    assert(BN._hasBuffer);
-    return new Buffer(this.toArray(endian, length));
+    assert(typeof BN._Buffer !== 'undefined');
+    return new BN._Buffer(this.toArray(endian, length));
   };
 
   BN.prototype.toArray = function toArray (endian, length) {
--- a/lib/bn.js
+++ b/lib/bn.js
@@ -51,10 +51,10 @@
   BN.BN = BN;
   BN.wordSize = 26;
 
+  var Buffer;
   try {
-    BN._Buffer = require('buf' + 'fer').Buffer;
+    Buffer = require('buf' + 'fer').Buffer;
   } catch (e) {
-    BN._Buffer = undefined;
   }
 
   BN.max = function max (left, right) {
@@ -502,8 +502,8 @@
   };
 
   BN.prototype.toBuffer = function toBuffer (endian, length) {
-    assert(typeof BN._Buffer !== 'undefined');
-    return new BN._Buffer(this.toArray(endian, length));
+    assert(typeof Buffer !== 'undefined');
+    return new Buffer(this.toArray(endian, length));
   };
 
   BN.prototype.toArray = function toArray (endian, length) {
--- a/lib/bn.js
+++ b/lib/bn.js
@@ -495,6 +495,10 @@
     return this.toString(16);
   };
 
+  BN.prototype.toBuffer = function toBuffer (endian, length) {
+    return new Buffer(this.toArray(endian, length));
+  };
+
   BN.prototype.toArray = function toArray (endian, length) {
     var byteLength = this.byteLength();
     var reqLength = length || Math.max(1, byteLength);
--- a/lib/bn.js
+++ b/lib/bn.js
@@ -50,6 +50,7 @@
 
   BN.BN = BN;
   BN.wordSize = 26;
+  BN._hasBuffer = typeof Buffer !== 'undefined';
 
   BN.max = function max (left, right) {
     if (left.cmp(right) > 0) return left;
@@ -496,6 +497,7 @@
   };
 
   BN.prototype.toBuffer = function toBuffer (endian, length) {
+    assert(BN._hasBuffer);
     return new Buffer(this.toArray(endian, length));
   };
 
--- a/README.md
+++ b/README.md
@@ -60,6 +60,7 @@ either `le` (little-endian) or `be` (big-endian).
   pad to length, throwing if already exceeding
 * `a.toString(base, padding)` - convert to base-string and pad with zeroes
 * `a.toNumber()` - convert to Javascript Number (limited to 53 bits)
+* `a.toBuffer()` - convert to Node.js Buffer (if available)
 * `a.bitLength()` - get number of bits occupied
 * `a.zeroBits()` - return number of less-significant consequent zero bits
   (example: `1010000` has 4 zero bits)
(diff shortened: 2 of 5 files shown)


jprichardson:
:+1:

lib/bn.js
```diff
@@ -52,6 +52,7 @@ else
 
 BN.BN = BN;
 BN.wordSize = 26;
+BN._hasBuffer = typeof Buffer !== 'undefined';
```
indutny: How does it play with browserify? Will it include `Buffer` for browsers?

axic: Good catch, just checked and it does include the browserify version of Buffer.

However ignoring that module alleviates it: ```browserify --ignore buffer lib/bn.js -s BN -o bn-browser.js```

indutny: I'm not sure if it will work well when `bn.js` is used as dependency. Usually, I work-around such problems by doing `require('buf' + 'fer').Buffer` inside of `try { } catch {}`.

In this case, though, such check should happen lazily, IMO.

lib/bn.js
```diff
@@ -459,6 +460,11 @@ BN.prototype.toJSON = function toJSON() {
   return this.toString(16);
 };
 
+BN.prototype.toBuffer = function toBuffer(endian, length) {
```
axic: Sounds good, but Buffer is not an exact match. ```push``` and ```unshift``` are not available.

indutny: Yeah, it will require a bit of rewriting of code below, but I think it should be possible. Are you interested in working on this?

axic: I am not entirely sold that ```toArrayLike(Buffer, ...)``` would be handier to use than ```new Buffer(bn.toArray(...))```.

However BN itself could export an index method and therefore become indexable as an Array?

Or alternatively it could make sense to have a node-bn.js package which has these kind of extensions? Although toBuffer on its own might be too slim for that.

indutny: I didn't propose to make it an external API. Rather a common part for both `toArray` and `toBuffer`.

indutny: `toBuffer` seems to be very minor.

axic: Ah ok, I thought it is part of the dependency problem for browserify, because using it like the above solves that issue.

lib/bn.js
axic: Pushed an implementation based on the suggested. It seems to work with browserify, the tests and an actual web browser.

lib/bn.js
axic: I think implementing ```toArrayLike``` should be left as an exercise for a separate pull request, because it is about optimisation.

lib/bn.js
```diff
@@ -459,6 +465,11 @@ BN.prototype.toJSON = function toJSON() {
...
+  assert(typeof BN._Buffer !== 'undefined');
```
indutny: Is there any point in exposing `_Buffer` to users? What do you think about making at a context variable instead?

lib/bn.js
```diff
@@ -53,6 +53,12 @@ else
 BN.BN = BN;
 BN.wordSize = 26;
 
+try {
```
indutny: Like declaring variable here.

lib/bn.js
indutny: Sorry, but I disagree. Most of the functions here were created with the aim for performance, and introducing `toBuffer` means certain expectations from the API users. I would appreciate if we could meet this expectations and make it as fast as we can.

Thanks!

lib/bn.js
axic: What should be the variable name? Ideas:
- _Buffer or __Buffer
- nodeBuffer
- nodejsBuffer
- BufferCons
- BufferAPI

I think I would go with nodeBuffer or nodejsBuffer.

indutny: just `Buffer`, IMO.

lib/bn.js
```diff
@@ -53,6 +53,12 @@ else
...
+  BN._Buffer = require('buf' + 'fer').Buffer;
```
odeke-em: Just out of curiosity why
```javascript
require('buf' + 'fer')
```
instead of
```javascript
require('buffer');
```

indutny: Because of `browserify`.

odeke-em: Oh wow.

Kagami: Like I already mentioned in [PR to your other project](https://github.com/indutny/brorand/pull/3), webpack is (unfortunately) smart enough to break this hack.

dcousens:
Out of interest,  @indutny what are you trying to prevent?

lib/bn.js
dcousens: Out of interest,  @indutny what are you trying to prevent?

indutny: I don't want to bundle `buffer` when linking this module in browserify.

lib/bn.js
axic: I still believe this should be part of a separate pull request as it involves more changes.

The first step is https://github.com/indutny/bn.js/pull/81 to remove the push/unshift dependency. After that it is simple.

axic: Based on the merged #81, the following patch works fine:

```diff
diff --git a/lib/bn.js b/lib/bn.js
index b657b34..7787c04 100644
--- a/lib/bn.js
+++ b/lib/bn.js
@@ -467,17 +467,21 @@ BN.prototype.toJSON = function toJSON() {
 
 BN.prototype.toBuffer = function toBuffer(endian, length) {
   assert(typeof Buffer !== 'undefined');
-  return new Buffer(this.toArray(endian, length));
+  return this.toArrayLike(Buffer, endian, length);
 };
 
 BN.prototype.toArray = function toArray(endian, length) {
+  return this.toArrayLike(Array, endian, length);
+};
+
+BN.prototype.toArrayLike = function toArray(cons, endian, length) {
   var byteLength = this.byteLength();
   var reqLength = length || byteLength;
   assert(byteLength <= reqLength, 'byte array longer than desired length');
 
   this.strip();
   var littleEndian = endian === 'le';
-  var res = new Array(reqLength);
+  var res = new cons(reqLength);
 
   var q = this.clone();
   if (!littleEndian) {
```

axic:

lib/bn.js
axic: @indutny @Kagami I've came across this version of instructing browser of non inclusion (from here: https://github.com/jprichardson/secure-random/commit/bccdbd2004c0895f8a3a382db19f5aac7643a6a1), might worth a try?

It works by adding a section to `package.json`

indutny:
Hell yes, this looks great!

axic:
@indutny I did a rebase against master, should I force push it to this PR (or should I create a new PR) so you can merge it easily?

lib/bn.js
axic: Since `semistandard` is used now, the above will fail with: `bn.js:510:19: A constructor name should not start with a lowercase letter.` The parameter must be named `Cons` to make the `new` keyword happy.

indutny:
Force push should be fine, assuming that you didn't do merge commits in your branch.

axic:
Done, there should be no conflicts

indutny:
Hm... I thought that there was suggestion about using `browser` field in package.json. What do you think about this @axic ?

axic:
@indutny I suggested it, but there was no response

I tested it and works, at least the browserified output is the same size, but not an exact copy.

```js
},{"buffer":2}],2:[function(require,module,exports){
```

It includes an empty package named `buffer`. With the try/catch it works.

indutny:
Great, let's do it this way. Sorry for not responding!

axic:
Pushed the package.json way

indutny:
PR no longer applies... May I ask you to rebase?

axic:
@indutny github said there are no conflicts, but rebased anyway

indutny:
Landed, thank you! (Released too)

Issue #72, opened 2015-11-21
Feature request: bitwise negation


axic:
It would be nice to have a method for doing bitwise negation.

indutny:
Example?

axic:
```a.not()``` and ```a.inot()``` - bitwise negate (NOT) all bits (up to bitLength())

axic:
Perhaps it there should be two versions: ```a.not()``` would negative up to ```a.bitLength()```, but ```a.notn(b)``` would also include all the leading zeroes (and set them to 1), where the number of leading zeroes equal to ```b - a.bitLength()```.

Assuming a.bitLength() equals to the position of the most significant bit set.

indutny:
So basically `a.not().not()` will not be equal to `a`?

axic:
Correct, it wouldn't. Not sure ```a.not()``` is the right one to have. I am interested in having ```a.notn(b)```.

indutny:
@axic what is your use case?

axic:
My actual use case was https://github.com/indutny/bn.js/issues/73, but have done it differently since. I cannot think about any use case top of my head, but my hunch is it can be useful in the future.

That being said, I'm not interested in implementing it. Perhaps this issue can stay open until someone does it.

axic:
Meanwhile my API suggestion for this would be:
- ```inotn(width)```
- ```notn(width)```
- ```inot()``` - where bitLength() is used as width
- ```not()``` - where bitLength() is used as width

So that ```a.inotn(256).inotn(256) === a```, but ```a.inot().inot()``` might differ.

If it is an intent to avoid user error, ```inot()``` and ```not()``` should not be added.

I have also created sample implementations and will create pull requests.

axic:
PR https://github.com/indutny/bn.js/pull/77 was merged, this is done.

PR #77, opened 2015-12-01
Support bitwise NOT

axic:
Implements the interface discussed in https://github.com/indutny/bn.js/issues/72.

Note that I've kept the lazy shift based implementation as the first commit. The second commit using the data directly is twice as fast according to the benchmarks. (Created my own benchmarks based on the benchmarking tool in the tree.)

### DIFF
--- a/lib/bn.js
+++ b/lib/bn.js
@@ -748,16 +748,16 @@ BN.prototype.inotn = function inotn(width) {
 
   // Handle complete words
   for (var i = 0; i < bytesNeeded; i++) {
-    var tmp = this.words[i] & 0xfc000000;
-    this.words[i] = tmp | (~(this.words[i] & 0x3ffffff)) & 0x3ffffff;
+    this.words[i] = (this.words[i] & 0xfc000000) |
+                    ((~(this.words[i] & 0x3ffffff)) & 0x3ffffff);
   }
 
   // Handle the residue
   if (bitsLeft > 0) {
     var tmask = 0xfc000000 | ((0xffffffff >> bitsLeft) << bitsLeft);
     var bmask = 0x3ffffff >> (26 - bitsLeft);
-    var tmp = this.words[i] & tmask;
-    this.words[i] = tmp | (~(this.words[i] & bmask)) & bmask;
+    this.words[i] = (this.words[i] & tmask) |
+                    ((~(this.words[i] & bmask)) & bmask);
   }
 
   // And remove leading zeroes
--- a/lib/bn.js
+++ b/lib/bn.js
@@ -736,15 +736,18 @@ BN.prototype.inotn = function inotn(width) {
   assert(typeof width === 'number' && width >= 0);
 
   var bytesNeeded = Math.ceil(width / 26) | 0;
-  var bitsLeft = width;
+  var bitsLeft = width % 26;
 
   // Extend the buffer with leading zeroes
   while (this.length < bytesNeeded) {
     this.words[this.length++] = 0;
   }
 
+  if (bitsLeft > 0)
+    bytesNeeded--;
+
   // Handle complete words
-  for (var i = 0; i < bytesNeeded && bitsLeft >= 26; i++, bitsLeft -= 26) {
+  for (var i = 0; i < bytesNeeded; i++) {
     var tmp = this.words[i] & 0xfc000000;
     this.words[i] = tmp | (~(this.words[i] & 0x3ffffff)) & 0x3ffffff;
   }
--- a/lib/bn.js
+++ b/lib/bn.js
@@ -735,7 +735,7 @@ BN.prototype.uxor = function uxor(num) {
 BN.prototype.inotn = function inotn(width) {
   assert(typeof width === 'number' && width >= 0);
 
-  var bytesNeeded = width / 26;
+  var bytesNeeded = Math.ceil(width / 26) | 0;
 
   // Extend the buffer with leading zeroes
   while (this.length < bytesNeeded) {
--- a/lib/bn.js
+++ b/lib/bn.js
@@ -732,8 +732,41 @@ BN.prototype.uxor = function uxor(num) {
 
 
 // Not ``this`` with ``width`` bitwidth
+BN.prototype.inotn = function inotn(width) {
+  assert(typeof width === 'number' && width >= 0);
+
+  var bytesNeeded = width / 26;
+
+  // Extend the buffer with leading zeroes
+  while (this.length < bytesNeeded) {
+    this.words[this.length++] = 0;
+  }
+
+  for (var i = 0; i < bytesNeeded; i++) {
+    var tmask = 0xfc000000; // top field
+    var bmask = 0x3ffffff;  // bottom field
+
+    if (width < 26) {
+      // keep the bottom <width> bits
+      bmask >>= (26 - width);
+      // increase the top mask
+      tmask |= (tmask >> width) << width;
+    }
+    width -= 26;
+
+    // NOT the bottom field and leave the top intact
+    var tmp = this.words[i] & tmask;
+    this.words[i] = tmp | (~(this.words[i] & bmask)) & bmask;
+  }
+
+  // And remove leading zeroes
+  this.strip();
+
+  return this;
+};
+
 BN.prototype.notn = function notn(width) {
-  return new BN(1).iushln(width).isubn(1).isub(this);
+  return this.clone().inotn(width);
 };
 
 
--- a/lib/bn.js
+++ b/lib/bn.js
@@ -748,13 +748,12 @@ BN.prototype.inotn = function inotn(width) {
 
   // Handle complete words
   for (var i = 0; i < bytesNeeded; i++) {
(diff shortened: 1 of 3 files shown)


dcousens:
LGTM

lib/bn.js
```diff
@@ -731,6 +731,45 @@ BN.prototype.uxor = function uxor(num) {
...
+    }
+    width -= 26;
+
+    // NOT the bottom field and leave the top intact
+    var tmp = this.words[i] & tmask;
+    this.words[i] = tmp | (~(this.words[i] & bmask)) & bmask;
+  }
+
+  // And remove leading zeroes
+  this.strip();
```
indutny: Let's do `return this.strip();`

lib/bn.js
```diff
@@ -731,6 +731,45 @@ BN.prototype.uxor = function uxor(num) {
...
+  // Extend the buffer with leading zeroes
+  while (this.length < bytesNeeded) {
+    this.words[this.length++] = 0;
+  }
+
+  for (var i = 0; i < bytesNeeded; i++) {
+    var tmask = 0xfc000000; // top field
+    var bmask = 0x3ffffff;  // bottom field
+
+    if (width < 26) {
```
indutny: Do you think we could "un-peel" this condition out of the loop? It seems to me that the if's body is executed only at the very end of the loop.

lib/bn.js
```diff
@@ -731,6 +731,45 @@ BN.prototype.uxor = function uxor(num) {
 };
 
 
+// Not ``this`` with ``width`` bitwidth
+BN.prototype.inotn = function inotn(width) {
+  assert(typeof width === 'number' && width >= 0);
+
+  var bytesNeeded = width / 26;
```
indutny: `Math.ceil(width / 26)`, perhaps?

lib/bn.js
```diff
@@ -731,6 +731,45 @@ BN.prototype.uxor = function uxor(num) {
...
+    var tmask = 0xfc000000; // top field
+    var bmask = 0x3ffffff;  // bottom field
+
+    if (width < 26) {
+      // keep the bottom <width> bits
+      bmask >>= (26 - width);
+      // increase the top mask
+      tmask |= (tmask >> width) << width;
+    }
+    width -= 26;
```
indutny: I think we need a separate variable, V8 (and others) do not really work well when code modifies arguments.

lib/bn.js
axic: I had it there before but it doesn't really make a difference - even in the case of 27/26, 1<1.0384.. == 1<2. Added back to avoid issues with broken JS interpreters.

axic: It doesn't really make a difference - even with 1/26 it is 0.0384.. and that is bigger than 0. Added it back for broken JS interpreters which would evaluate 0<0.0384 as false.

indutny: The issue is that you comparing integers with a floating point number in a loop below, which is slower than comparing integer to integer (obviously).

dcousens: @indutny is `Math.ceil` even guaranteed to give back an integer?

indutny: Argh, you are totally right. We need `Math.ceil(num) | 0`.

axic:
Pushed the changes. Perhaps they will be squashed together when cherry picked, but I think keeping at least b71c485 and 2a4a25b (with or without the rest squashed into this) as commits would be helpful.

axic:
I am not entirely sure about different JS interpreters and how they adhere to data precedence/access, but the below simpler version might be suitable for all?

```
@@ -745,16 +745,14 @@ BN.prototype.inotn = function inotn(width) {
 
   // Handle complete words
   for (var i = 0; i < bytesNeeded && bitsLeft >= 26; i++, bitsLeft -= 26) {
-    var tmp = this.words[i] & 0xfc000000;
-    this.words[i] = tmp | (~(this.words[i] & 0x3ffffff)) & 0x3ffffff;
+    this.words[i] = (this.words[i] & 0xfc000000) | (~(this.words[i] & 0x3ffffff)) & 0x3ffffff;
   }
 
   // Handle the residue
   if (bitsLeft > 0) {
     var tmask = 0xfc000000 | ((0xffffffff >> bitsLeft) << bitsLeft);
     var bmask = 0x3ffffff >> (26 - bitsLeft);
-    var tmp = this.words[i] & tmask;
-    this.words[i] = tmp | (~(this.words[i] & bmask)) & bmask;
+    this.words[i] = (this.words[i] & tmask) | (~(this.words[i] & bmask)) & bmask;
   }
 
   // And remove leading zeroes
```

indutny:
@axic there is no such differences between different JS compilers, because they are compliant to specs (and because the web will broke terribly if such basis things won't work consistently)

axic:
So you want that last piece as a commit then? It passes the tests.

lib/bn.js
```diff
@@ -731,6 +731,41 @@ BN.prototype.uxor = function uxor(num) {
...
+  var bytesNeeded = Math.ceil(width / 26) | 0;
+  var bitsLeft = width;
+
+  // Extend the buffer with leading zeroes
+  while (this.length < bytesNeeded) {
+    this.words[this.length++] = 0;
+  }
+
+  // Handle complete words
+  for (var i = 0; i < bytesNeeded && bitsLeft >= 26; i++, bitsLeft -= 26) {
```
indutny: I think `i < bytesNeeded` and `bitsLeft >= 26` are the same thing, doesn't look like we need both of them. Or am I wrong?

lib/bn.js
```diff
@@ -731,6 +731,41 @@ BN.prototype.uxor = function uxor(num) {
...
+    var tmp = this.words[i] & 0xfc000000;
+    this.words[i] = tmp | (~(this.words[i] & 0x3ffffff)) & 0x3ffffff;
+  }
+
+  // Handle the residue
+  if (bitsLeft > 0) {
```
indutny: There is no need in calculating residue in a loop. You may want to use `%` instead, and loop over `i`.

lib/bn.js
axic: Yes, ```i < bytesNeeded``` can be removed.

lib/bn.js
axic: Either you take track of the bits left in the loop and stop when appropriate; or calculate the ending point before, or simply just leave the top part always to the residue loop.

```
@@ -736,7 +736,7 @@ BN.prototype.inotn = function inotn(width) {
   assert(typeof width === 'number' && width >= 0);
 
   var bytesNeeded = Math.ceil(width / 26) | 0;
-  var bitsLeft = width;
+  var bitsLeft = width % 26;
 
   // Extend the buffer with leading zeroes
   while (this.length < bytesNeeded) {
@@ -744,7 +744,7 @@ BN.prototype.inotn = function inotn(width) {
   }
 
   // Handle complete words
-  for (var i = 0; i < bytesNeeded && bitsLeft >= 26; i++, bitsLeft -= 26) {
+  for (var i = 0; i < bytesNeeded - 1; i++) {
     var tmp = this.words[i] & 0xfc000000;
     this.words[i] = tmp | (~(this.words[i] & 0x3ffffff)) & 0x3ffffff;
   }
```

To calculate actual end point, replace in the above:
```
+  for (var i = 0; i < bytesNeeded - ((bitsLeft > 0) ? 1 : 0); i++) {
```

Which way you prefer?

indutny: I think you could decrement `bytesNeeded` before the loop start if `bitsLeft !== 0`.

axic: Added this version. I would appreciate if you could just let me know upfront the way you want it so we don't go in circles :)

axic:
@indutny is there anything else to be changed or you are happy with current version?

lib/bn.js
```diff
@@ -731,6 +731,44 @@ BN.prototype.uxor = function uxor(num) {
...
+  while (this.length < bytesNeeded) {
+    this.words[this.length++] = 0;
+  }
+
+  if (bitsLeft > 0)
+    bytesNeeded--;
+
+  // Handle complete words
+  for (var i = 0; i < bytesNeeded; i++) {
+    this.words[i] = (this.words[i] & 0xfc000000) |
```
indutny: I have a question. What is this `0xfc000000` mask for? All values in `this.words` are always `>= 0` and `<=0x3ffffff`.

axic: I didn't do a full code review to see if anything is done to the upper bits (masked by ```0xfc000000```). I assumed there might be something there and to be on the safe side I decided to keep it as is while NOTing the lower bits.

If you say the top is never used (apart from being an overhead space for in-place operations), that part can be discarded.

IIRC I had weird issues by just doing a NOT on the whole words[i] as obviously that was my first try :)

indutny: Of course. Thank you very much, and sorry for long comments thread on this PR! ;)

indutny: And you don't need top bits indeed, please remove them.

axic: Discarding the top simplifies it a lot 👌

axic:
Is the last version looking good then? I think it is quite clear now.

indutny:
LGTM

indutny:
Landed and released ,thank you

PR #78, opened 2015-12-01
bn: introduce '.isZero()'

axic:
As the title says.

The actual code is taken from here:
```
BN.prototype._normSign = function _normSign() {
  // -0 = 0
  if (this.length === 1 && this.words[0] === 0)
    this.negative = 0;
  return this;
};
```

The aim is to replace all the various suboptimal checks for zero (like ```if (this.cmpn(0) === 0)```) with this.

(Of course I am doing zero-checks in my tools too, so it is handy there too)


indutny:
LGTM

indutny:
Landed, thank you!

PR #81, opened 2015-12-10
bn: optimise toArray to not use push/unshift

axic:
This change is required to implement ```toArrayLike``` functionality requested in https://github.com/indutny/bn.js/pull/76

### DIFF
--- a/lib/bn.js
+++ b/lib/bn.js
@@ -460,19 +460,24 @@ BN.prototype.toJSON = function toJSON() {
 };
 
 BN.prototype.toArray = function toArray(endian, length) {
+  var byteLength = this.byteLength();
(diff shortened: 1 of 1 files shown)


lib/bn.js
```diff
@@ -460,19 +460,24 @@ BN.prototype.toJSON = function toJSON() {
 };
 
 BN.prototype.toArray = function toArray(endian, length) {
+  var byteLength = this.byteLength();
+  length = length || byteLength;
```
indutny: Modifying arguments is bad idea, kills the performance.

axic: Fixed in 887e11d

indutny:
Landed, thank you!

Issue #82, opened 2015-12-10
handling zero on sub bug


wanderer:
This is not what I expected
`new BN(0).divn(100000).subn(2)`
`<BN: 3fffffe>`

I would think it would just be `-2`

indutny:
I'm sorry, but `subn` does not support sign change :( Can't say if it is intended or not, perhaps just a documentation bug.

Though, if you have an idea on how to add support for sign change for both (`addn` and `subn`) whilst keeping the speed of current implementation - please let me know!

axic:
It should throw an assert in that case as the rest of the functions do - shouldnt it?

wanderer:
should be fixed now

PR #83, opened 2015-12-11
fixed subi bug

wanderer:
#82

### DIFF
--- a/lib/bn.js
+++ b/lib/bn.js
@@ -2715,10 +2715,15 @@ BN.prototype.isubn = function isubn(num) {
 
   this.words[0] -= num;
 
-  // Carry
-  for (var i = 0; i < this.length && this.words[i] < 0; i++) {
-    this.words[i] += 0x4000000;
-    this.words[i + 1] -= 1;
+  if (this.length === 1 && this.words[0] < 0) {
+    this.words[0] = -this.words[0];
+    this.negative = 1;
+  } else {
+    // Carry
+    for (var i = 0; i < this.length && this.words[i] < 0; i++) {
+      this.words[i] += 0x4000000;
+      this.words[i + 1] -= 1;
+    }
   }
 
   return this.strip();
--- a/test/arithmetic-test.js
+++ b/test/arithmetic-test.js
@@ -136,6 +136,16 @@ describe('BN.js/Arithmetic', function() {
       assert.equal(a.negative, 0);
       assert.equal(a.toString(), '100');
     });
+
+    it('should change sign on small numbers at 0', function() {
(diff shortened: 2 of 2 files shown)


lib/bn.js
```diff
@@ -2715,10 +2715,15 @@ BN.prototype.isubn = function isubn(num) {
 
   this.words[0] -= num;
 
-  // Carry
-  for (var i = 0; i < this.length && this.words[i] < 0; i++) {
-    this.words[i] += 0x4000000;
-    this.words[i + 1] -= 1;
+  if (this.words[1] === undefined && this.words[0] < 0) {
```
indutny: `this.words[1]` is a total garbage (could be a number), if `length` is `1`. I would rather rely on `this.length` instead.

wanderer: @indutny fixed

lib/bn.js
```diff
@@ -2715,10 +2715,15 @@ BN.prototype.isubn = function isubn(num) {
 
   this.words[0] -= num;
 
-  // Carry
-  for (var i = 0; i < this.length && this.words[i] < 0; i++) {
-    this.words[i] += 0x4000000;
-    this.words[i + 1] -= 1;
+  if (this.words.length === 1 && this.words[0] < 0) {
```
indutny: `this.words.length` is not representative either, sorry! Please consider using `this.length` instead.

test/arithmetic-test.js
```diff
@@ -136,6 +136,11 @@ describe('BN.js/Arithmetic', function() {
       assert.equal(a.negative, 0);
       assert.equal(a.toString(), '100');
     });
+
+    it('should change sign on small numbers', function() {
+      var a = new BN(0).subn(2)
```
indutny: I would suggest adding one more test, like `1 - 2`

indutny:
Looks really great, thank you!

test/arithmetic-test.js
wanderer: ok done!

lib/bn.js
wanderer: fixed!

wanderer: fixed!

test/arithmetic-test.js
```diff
@@ -136,6 +136,16 @@ describe('BN.js/Arithmetic', function() {
...
+      assert.equal(a.toString(), '-2');
+    });
+
+    it('should change sign on small numbers', function() {
```
indutny: This should have different name.

wanderer: done!

indutny:
Thank you, landed and released!

PR #84, opened 2015-12-14
Add tests for cmp/cmpn, a tiny optim to cmp & update README

axic:


indutny:
Just curious is there any measurable performance difference with this approach?

axic:
I didn't measure it and I would guess it would be light on its own, but .cmpn is used a lot so it might actually add up.  In any case it feels cleaner.

indutny:
@axic it doesn't feel that cleaner to me, honestly saying. May I ask you to measure the speed difference one some artificial benchmark? Maybe just `cmpn` in a loop with big enough number of iterations?

axic:
Actually I planned to submit another follow up PR to replace the mask with assert so the user will be aware of the limitation. It makes more sense to have the code piece in that branch for that also.

```assert(num <= 0x3ffffff, 'Number is too big');```

The test results (based on benchmark/, where both the old, this proposal and assert is included):
```
Benchmarking: cmp
bn.js#cmp x 56,083,525 ops/sec ±3.87% (9 runs sampled)
old bn.js#cmp x 54,656,053 ops/sec ±2.66% (9 runs sampled)
assert bn.js#cmp x 56,243,928 ops/sec ±3.29% (9 runs sampled)
------------------------
Fastest is assert bn.js#cmp,bn.js#cmp,old bn.js#cmp
========================

Benchmarking: cmp2
bn.js#cmp2 x 44,763,118 ops/sec ±2.74% (9 runs sampled)
old bn.js#cmp2 x 43,464,701 ops/sec ±2.18% (9 runs sampled)
assert bn.js#cmp2 x 45,479,641 ops/sec ±1.76% (9 runs sampled)
------------------------
Fastest is assert bn.js#cmp2,bn.js#cmp2
========================
```

cmp is bn('fffffffffffffffffffffffffffffffffffffffffffffffffffffffffffffffe', 16) vs. -42
cmp2 is bn('fffffe', 16) vs. -42

axic:
@indutny well actually I figured since it is in the benchmarks, see the assert change in https://github.com/indutny/bn.js/commit/911e93653d0462f08757f2fff4bf2a1a5df31727

indutny:
I'm convinced. Thank you! LGTM

indutny:
Landed!

axic:
Great! Thank you!

PR #86, opened 2015-12-15
bn: replace cmpn() with isZero() as it is faster

axic:
As promised in #78 here is the update to make use of isZero and speed up other methods.


indutny:
LGTM. Landed and released, thank you!

PR #87, opened 2015-12-16
benchmark: use bignum 0.11.0, which works on OSX too

axic:


indutny:
LGTM, thank you!

indutny:
Landed.

PR #88, opened 2015-12-16
benchmark: support github.com/MikeMcl/bignumber.js

axic:


indutny:
It seems like `bignumber#sub:` is crashing. Apart from this, do you think adding this library to benchmarks will provide any value? It doesn't seem like any of covered functions are faster in `bignumber`.

axic:
Thankfully it is not anywhere near to bn.js. I have a couple of projects I would like to convince to switch to bn.js form bignumber and having benchmarks certainly helps.

Didin't notice .sub not working, strange.

axic:
@indutny rebased against master and fixed .sub. All the methods in bignumber.js are faster than silentmatt-biginteger.

indutny:
Ah, closing this :)

axic:
To the defend the point of this, here's a comparison of *Github stars*, *npm dependents* and *npm downloads last month* of each of the bignum libraries used **currently** in benchmarks:
- `bignum` (node-bignum), 211 stars, 49 deps, 51,687 downloads
- `sjcl` (eccjs), 41 stars, 5 deps, 86 downloads
- `bigi` 8 stars, 39 deps, 6171 downloads
- `js-big-integer` 10 stars, 9 deps, 50 downloads
- `biginteger` 133 stars, 2 deps, 93 downloads

How bignumber compares:
- `bignumber.js` 667 stars, >50 deps, 333,311 downloads

Note: the package `sjcl` should be used instead of `eccjs` as it is more update and has more users: 3113 stars, 42 deps, 5396 downloads

My points is that benchmarks should be made against:
a) better performing libraries;
b) widely used libraries.

indutny:
Alright, let's throw away `biginteger` then, and replace it with `bignumber.js`.

indutny:
Also good point about `sjcl`.

Issue #90, opened 2015-12-19
Bug in parseBase for numbers containing a dot ('0.0079')


axic:
``` js
new BN('0.0079', 10) // <BN: ffffb22f>
new BN('0.02', 10) // <BN: ffffff3a>
```

`_init`, `parseBase` or `_parseBase` will not care if a dot is included in a base-10 input, will process it as `('.' - 48) & 0xf`.

I think it should just reject such input?

axic:
The same applies to ```parseHex``` (as basically it is the same code as ```parseBase```, with one exception: range is limited to f/F).

```js
new BN('1234', 16) // <BN: 1234>
new BN('12.34', 16) // <BN: 12e34>
```

indutny:
I know that it should throw, but it doesn't right now. The thing is that this is an big __integer__ library, so you should not really pass data like this to it.

axic:
I do know that and I don't pass those kind of numbers. However, users of my library are. I can either do pre-processing there which is quite big of an overhead. I would prefer bn.js to throw, I have the PR ready, will submit it soon - it is quite small.

axic:
A note on the *integer* side, the library's name is short for big number and doesn't specifically mentions anywhere it is integers and not decimals. I know that, but it is not clear from the first sight at the README, maybe it should be updated.

indutny:
Yeah, I totally agree. It should be updated. Thanks!

fanatid:
`new BN(4.2)` return `<BN: 4>`
also should be assert here

dcousens:
A note was added to the README at some point: `Note: decimals are not supported in this library.`

indutny:
It throws now.

PR #91, opened 2015-12-19
Reject decimal input

axic:
As discussed in https://github.com/indutny/bn.js/issues/90

### DIFF
--- a/lib/bn.js
+++ b/lib/bn.js
@@ -233,16 +233,17 @@
 
       r *= mul;
 
-      // 'a'
-      if (c >= 49) {
+      // 'a' - 'z'
+      if (c >= 49 && c <= 74) {
         r += c - 49 + 0xa;
 
-      // 'A'
-      } else if (c >= 17) {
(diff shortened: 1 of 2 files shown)


indutny:
Does it affect benchmarks?

axic:
Seems like my comments didn't get posted here back in December. Unfortunately it does affect benchmarks so I was thinking about others ways to do it. Perhaps the best would be having two input functions: a *safe* (for "user input") and an *unsafe* one (for speed, e.g. the current implementation).

Any ideas about an API for the above?

indutny:
I'm not sure about it myself. Perhaps, `BN.checkedFrom(string, 'hex')`?

indutny:
Preemptively added `semver-major` tag per your request.

fanatid:
Why not land this PR in `4.x`?
@axic I think only *safe* should be here

indutny:
Landed test in 9173bc2, the issue was fixed in different commit. Thank you!

PR #97, opened 2016-01-17
redNeg and zero

fanatid:


indutny:
Description of issue, test case?

fanatid:
When number is zero `redNeg` should return zero, not `this.m` value

indutny:
Landed in 0d55721, please put the more detailed description of the issue in commit log in the future. Thank you!

PR #98, opened 2016-01-17
bn: fix Red.isqr

fanatid:
In #65 @alexeykudinkin fixed `isqr`, but Red.isqr have same bug.


indutny:
Test case! ;)

indutny:
Landed, thank you!

PR #99, opened 2016-01-18
bn: Fix sign in comb10MulTo

fanatid:
@indutny we forgot about sign in `comb10MulTo`


indutny:
Ouch!

indutny:
@fanatid do you have a regression test for this? May I ask you to include one in this PR?

indutny:
Landed, thank you!

Issue #96, opened 2016-01-17
mod.neg in divmod


fanatid:
https://github.com/indutny/bn.js/blob/c9d5ab8106d27c95c0e00c5651603bd35b279938/lib/bn.js#L2294
```js
if (mode !== 'div') {
  mod = res.mod.neg();
  if (positive && mod.neg) {
    mod = mod.add(num);
  }
}
```
`neg`? Will be always `true`, because `neg` is function.

indutny:
I guess it should be `sign`. Thank you for catching this!

Could you be interested in submitting a PR with a fix and test case for this?

indutny:
Or `negative`, whatever we have now.

axic:
@fanatid there is a second occurrence down the line: https://github.com/indutny/bn.js/blob/c9d5ab8106d27c95c0e00c5651603bd35b279938/lib/bn.js#L2320

axic:
Also shouldn't it be changed to `mod.iadd(num)`?

fanatid:
@axic thanks for note! yep, mod.iadd(num) is better, I think we also can change .neg() to .negative ^= 1

PR #100, opened 2016-01-18
bn: fix BN.divmod

fanatid:
See #96 for details

### DIFF
--- a/lib/bn.js
+++ b/lib/bn.js
@@ -623,15 +623,14 @@
 
   // Return negative clone of `this`
   BN.prototype.neg = function neg () {
-    if (this.isZero()) return this.clone();
-
-    var r = this.clone();
-    r.negative = this.negative ^ 1;
-    return r;
+    return this.clone().ineg();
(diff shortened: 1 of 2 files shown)


fanatid:
`BN.ineg` also was fixed.

indutny:
Needs test case. ;)

lib/bn.js
```diff
@@ -2281,51 +2280,45 @@
...
       if (mode !== 'mod') {
-        div = res.div.neg();
+        res.div.ineg();
       }
 
       if (mode !== 'div') {
-        mod = res.mod.neg();
-        if (positive && mod.neg) {
-          mod = mod.add(num);
+        res.mod.ineg();
```
indutny: I don't think that this is correct, `res.mod` could be `=== this`.

fanatid: You right, I forgot about this! Updated.

indutny:

indutny:

Issue #101, opened 2016-01-23
redSqrt error


fanatid:
On testing https://github.com/cryptocoinjs/secp256k1-node/pull/57 [Martin Becze](http://github.com/wanderer/) found that [elliptic](http://github.com/indutny/elliptic) and secp256k1-node embedded pure js implementation have error on recover operation.
* Message: f75c6b18a72fabc0f0b888c3da58e004f0af1fe14f7ca5d8c897fe164925d5e9
* Signature: fffffffffffffffffffffffffffffffebaaedce6af48a03bbfd25e8cd03641408887321be575c8095f789dd4c743dfe42c1820f9231f98a962b210e3ac2452a3
* Recovery ID: 0

But the real problem with [redSqrt](https://github.com/indutny/bn.js/blob/e82b88e2f11fe22c75f62b434d2f841e7fd619c6/lib/bn.js#L2805):
```js
var BN = require('bn.js')
var n = new BN('16e5f9d306371e9b876f04025fb8c8ed10f8b8864119a149803357e77bcdd3b1', 'hex').toRed(BN.red('k256'))
console.log(n.redSqr().redSqr().cmp(n))
```
give 1 instead 0

It's can be easily solving in bn.js by comparing result with [redSqr](https://github.com/indutny/bn.js/blob/e82b88e2f11fe22c75f62b434d2f841e7fd619c6/lib/bn.js#L2792) and throwing error... but how resolve it in elliptic context?
@indutny any ideas?

cc: @wanderer

indutny:
I'm not quite sure what you meant @fanatid . This sample: https://gist.github.com/indutny/38322b0b42eccd94d4e4 shows that red is producing the same number as if it would be computed without it... Do you think this is a bug in bn.js?

Where is the problem and what is expected result?

fanatid:
Oh, sorry, I skipped **t** in code... right version:
```js
var BN = require('bn.js')
var n = new BN('16e5f9d306371e9b876f04025fb8c8ed10f8b8864119a149803357e77bcdd3b1', 'hex').toRed(BN.red('k256'))
console.log(n.redSqrt().redSqr().cmp(n))
```

indutny:
Oh, this changes everything!

indutny:
@fanatid it looks like this number may not have square root module secp256k1. Is it expected to have recovery id equal to `0`? Or is `0` the result that it reports?

May I ask you to help me and submit the test case for elliptic?

fanatid:
[...]

PR #102, opened 2016-01-24
tests: redSqrt with quadratic nonresidue

fanatid:
Details: #101

### DIFF
--- a/test/red-test.js
+++ b/test/red-test.js
@@ -191,6 +191,14 @@ describe('BN.js/Reduction context', function () {
     });
   });
 
+  it('redSqrt for quadratic nonresidue', function () {
+    var red = BN.red('k256');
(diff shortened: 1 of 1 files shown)


indutny:
Heh... Looks like we did not understand each other after all. Sorry for this!

I will try to figure out test case that uses `elliptic` instead of raw-`bn.js` myself.

fanatid:
Oh, just now recognize that you write elliptic in 101 :smile:

axic:
@fanatid I think it probably makes sense having that test here too? After all it's `bn.js` methods :)

fanatid:
ok, branch is restored now, Fedor can fetch and make cherry-pick as usual :)

indutny:
@axic idk, I don't really want this method to throw, at least not right now. Maybe in next version?

PR #105, opened 2016-01-24
readme: document strip() and toJSON()

axic:


indutny:
`strip()` should not be documented, IMO. I can't imagine a use case for it, as all operation should strip numbers automatically.

axic:
Shouldn't `strip` be underscore prefixed (as the other internal methods) in that case?

indutny:
This is a good idea.

axic:
@indutny created a separate PR for strip and updated this one with a new patch: to document toJSON and make the readme clearer about the other export methods

indutny:

PR #106, opened 2016-01-24
bn: add lt/gt/eq shorthands for comparison

axic:
Adds lt/gt/eq (and n-postfixed versions) for a quick and user friendly helper. I still expect speed sensitive (and internal) methods to rely on `cmp`.

In terms of size: 40722 vs 41242 bytes when minified.


jprichardson:
I love this since other big integer / big number libraries support these methods.

indutny:
Alright, let's do it!

indutny:

PR #108, opened 2016-01-25
Add faster path for small numbers in .toNumber()

axic:


lib/bn.js
```diff
@@ -476,8 +476,14 @@
   };
 
   BN.prototype.toNumber = function toNumber () {
-    assert(this.bitLength() <= 53, 'Number can only safely store up to 53 bits');
-    return parseInt(this.toString(), 10);
+    var length = this.bitLength();
+    assert(length <= 53, 'Number can only safely store up to 53 bits');
+    if (length <= 26) {
+      var ret = this.words[0] & 0x3ffffff;
```
indutny: This `&` should not be necessary

lib/bn.js
```diff
@@ -476,8 +476,14 @@
...
+      return (this.negative !== 0) ? (ret * -1) : ret;
+    } else {
+      return parseInt(this.toString(), 10);
```
indutny: I wonder if it should do something like `this.words[1] * 0x4000000 + this.words[0]` here. Should speed up things significantly

axic: I was thinking about that, but that would mean just limiting toNumber to 52 bits instead of 53 and be done with it.

indutny: Ah, right! Maybe three cases then?

axic:
@indutny added the fast path for all 3 cases

indutny:
Thanks!

lib/bn.js
```diff
@@ -476,8 +476,19 @@
...
+      ret = this.words[0];
+    } else if (length <= 52) {
+      ret = (this.words[1] * 0x4000000) + this.words[0];
+    } else if (length === 53) {
+      // NOTE: at this stage it is known that the top bit is set
+      ret = 0x10000000000000 + (this.words[1] * 0x4000000) + this.words[0];
+    } else {
+      assert(false, 'Number can only safely store up to 53 bits');
+    }
+    return (this.negative !== 0) ? (ret * -1) : ret;
```
indutny: Heh, `-ret`, perhaps?

axic: I guess that works too :)

Done.

indutny:

axic:
Thanks!

PR #109, opened 2016-01-25
Add helper .isBN() to check if the supplied object is a BN.js instance

axic:
The logic is the same as in the constructor:
```js
  function BN (number, base, endian) {
    // May be `new BN(bn)` ?
    if (number !== null &&
      typeof number === 'object' &&
      Array.isArray(number.words)) {
      return number;
    }
    ...
```

It also could replace this part in the constructor, albeit I assume that could lead to a minor speed penalty in cases

### DIFF
--- a/lib/bn.js
+++ b/lib/bn.js
@@ -58,8 +58,8 @@
   }
 
   BN.isBN = function isBN (num) {
-    return (typeof num === 'object') &&
-      Array.isArray(num.words);
+    return (num !== null) && (typeof num === 'object') &&
+      (num.constructor.name === 'BN') && Array.isArray(num.words);
   };
 
   BN.max = function max (left, right) {
--- a/lib/bn.js
+++ b/lib/bn.js
@@ -57,6 +57,11 @@
   } catch (e) {
   }
 
+  BN.isBN = function isBN (num) {
+    return (typeof num === 'object') &&
+      Array.isArray(num.words);
+  };
+
   BN.max = function max (left, right) {
     if (left.cmp(right) > 0) return left;
     return right;
--- a/lib/bn.js
+++ b/lib/bn.js
@@ -19,10 +19,7 @@
   // BN
 
   function BN (number, base, endian) {
-    // May be `new BN(bn)` ?
-    if (number !== null &&
-      typeof number === 'object' &&
-      Array.isArray(number.words)) {
+    if (BN.isBN(number)) {
       return number;
     }
 
--- a/lib/bn.js
+++ b/lib/bn.js
@@ -55,8 +55,8 @@
   }
 
   BN.isBN = function isBN (num) {
-    return (num !== null) && (typeof num === 'object') &&
-      (num.constructor.name === 'BN') && Array.isArray(num.words);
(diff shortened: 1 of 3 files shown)


jprichardson:
I like this as well. `instanceof` checks can be buggy if the app has dependencies with different versions of `BN`. Although, I'd suggest using a field like `__bn = true` along with the aforementioned code. Then over time, as people upgrade their `BN` versions, we wouldn't have to rely upon duck typing.

axic:
@jprichardson there's also the line ```BN.BN = BN;``` which could be used for checks?

indutny:
@jprichardson `__bn` won't work with existing old versions, though :) Perhaps, `n.constructor.name === 'BN' && Array.isArray(n.words)`

jprichardson:
> Perhaps, n.constructor.name === 'BN' && Array.isArray(n.words)

:+1:

axic:
Would `typeof num === 'object'` still be needed or only going for the above?

axic:
Are we going with this final line?

`(typeof num === 'object') && (n.constructor.name === 'BN') && Array.isArray(n.words)`

indutny:
@axic `typeof null === 'object'` 🙁

axic:
So check for both then? 

`(num !== null) && (typeof num === 'object') && (n.constructor.name === 'BN') && Array.isArray(n.words)`

And shouldn't the constructor be updated with the check for `constructor.name` ?

indutny:
I think it should just use `BN.isBN` there.

axic:
Are there any more changes needed for this?

axic:

lib/bn.js
```diff
@@ -57,6 +54,11 @@
   } catch (e) {
   }
 
+  BN.isBN = function isBN (num) {
+    return (num !== null) && (typeof num === 'object') &&
```
indutny: No need in any of these parens.

test/utils-test.js
```diff
@@ -328,4 +328,15 @@ describe('BN.js/Utils', function () {
         '8000000000000000000000000000000000000000000000000000000000000000');
     });
   });
+
+  describe('.isBN', function () {
+    it('should return true for BN', function () {
+      assert.equal(BN.isBN(new BN()), true);
+    });
```
indutny: Please add newline here.

indutny:
Two comments, otherwise LGTM!

test/utils-test.js
axic: Between the two `it` blocks?

lib/bn.js
axic: For only these two or the next one as well?

indutny: For all!

test/utils-test.js
indutny: yep.

lib/bn.js
indutny: :wink:

axic:
@indutny done!

indutny:

axic:
Thanks!

Issue #110, opened 2016-01-25
GCD of negative number + zero


axic:
In the tests we have ```assert.equal(new BN(-18).gcd(new BN(12)).toString(16), '6');``` thus allowing negative numbers as an input.

For ```new BN(-18).gcd(new BN(0))``` we get -18, where GCD should be a positive number?

The same applies to ```new BN(0).gcd(new BN(-18))```.

If allowing gcd over a negative number is desired, the simple fix is to call `iabs()` here:
```js
  BN.prototype.gcd = function gcd (num) {
    if (this.isZero()) return num.clone();
    if (num.isZero()) return this.clone();
  ...
```

indutny:
Sounds like a bug to me.

axic:
This is fixed now

PR #111, opened 2016-01-25
tests: add tests for negative numbers to gcd/egcd/min/max

axic:
This has two failing tests and assumes https://github.com/indutny/bn.js/issues/110 is a bug and will be fixed.


indutny:
I don't really enjoy having failing tests in master, but I would appreciate having fix for those! ;)

axic:
Merge https://github.com/indutny/bn.js/pull/113 first and it will not be failing :)

indutny:
Thank you, landed!

PR #115, opened 2016-01-28
remove blacklisting of buffer in browser

wanderer:
The removed json mess up browserify by not letting it shim Buffer.

### DIFF
--- a/package.json
+++ b/package.json
@@ -29,8 +29,5 @@
     "istanbul": "^0.3.5",
     "mocha": "^2.1.0",
     "semistandard": "^7.0.4"
-  },
-  "browser": {
-    "buffer": false
   }
 }


indutny:
It should not shim Buffer in this library, right? Or is it messing up with other libraries?

wanderer:
It will mess up other libraries that do want to shim the Buffer. A top level project can decided to not shim the Buffer but if you disallow it here, dependents cannot override `Buffer: false`

indutny:
Shit, this sounds like an awful browserify bug. Have you already submitted the issue to it's repo?

axic:
Still the original way of Buffer require should solve the problem. Just reverse this: https://github.com/axic/bn.js/commit/815ba2bdcfe63e3d893653920470800c4549cead

wanderer:
@indutny I didn't think it was a bug. What is the rationale for disabling the buffer?

indutny:
The rationale is to minimize library size. There is a `toBuffer` method that is provided only for node.js.

wanderer:
>It will mess up other libraries that do want to shim the Buffer. A top level project can decided to not shim the Buffer but if you disallow it here, dependents cannot override Buffer: false

So it doesn't disable Buffer for all the whole project, just bn.js. Even if Buffer is included elsewhere [this check](https://github.com/indutny/bn.js/blob/master/lib/bn.js#L505) will always fail since browserify wraps each module in a IIFE

indutny:
@wanderer I think you forgot https://github.com/indutny/bn.js/blob/master/lib/bn.js#L56

felix:
This still seems broken for me using browserify. The JSON revert is good but the require() call only works if it is:

`    Buffer = require('buffer').Buffer;`

instead of:

`    Buffer = require('buf' + 'fer').Buffer;`

PR #116, opened 2016-01-29
bn: fix .toBuffer() inclusion and revert to old way (fixes #115)

axic:
This is a better fix to the problem reported in #115


wanderer:
:+1:

indutny:

PR #117, opened 2016-02-01
bn: make .strip() an internal method (see #105)

axic:


indutny:
Landed in e88e7b7, thank you!

PR #121, opened 2016-02-04
bn: introduce in-place pow, .ipow()

axic:
This changes pow() to be in-place and adds the usual wrapper.

Also added a test case for the special 0 power case.

### DIFF
--- a/README.md
+++ b/README.md
@@ -88,7 +88,7 @@ either `le` (little-endian) or `be` (big-endian).
 * `a.sub(b)` - subtraction (`i`, `n`, `in`)
 * `a.mul(b)` - multiply (`i`, `n`, `in`)
 * `a.sqr()` - square (`i`)
-* `a.pow(b)` - raise `a` to the power of `b`
+* `a.pow(b)` - raise `a` to the power of `b` (`i`)
 * `a.div(b)` - divide (`divn`, `idivn`)
 * `a.mod(b)` - reduct (`u`, `n`) (but no `umodn`)
 * `a.divRound(b)` - rounded division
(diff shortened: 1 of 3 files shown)


indutny:
Benchmarks? Last time I checked it - in-place multiplication was a bit slower than regular one. This is why I haven't used it in elliptic.

axic:
Perhaps the benchmarks aren't the best. The code is:
```js
var x = new bn(13);

add('pow', {
  'bn.js pow': function (fixture) {
    fixture.a1.clone().pow(x);
  },
  'bn.js ipow': function (fixture) {
    fixture.a1.clone().ipow(x);
  },
```

The results are:
```
new:
bn.js#pow x 7,371 ops/sec ±1.31% (9 runs sampled)
bn.js ipow#pow x 7,072 ops/sec ±4.80% (7 runs sampled)
bn.js#pow x 6,836 ops/sec ±7.52% (7 runs sampled)
bn.js ipow#pow x 6,955 ops/sec ±4.15% (7 runs sampled)
bn.js#pow x 7,504 ops/sec ±3.62% (9 runs sampled)
bn.js ipow#pow x 7,451 ops/sec ±2.85% (9 runs sampled)

old:
bn.js#pow x 7,475 ops/sec ±3.68% (9 runs sampled)
bn.js#pow x 6,838 ops/sec ±5.94% (9 runs sampled)
bn.js#pow x 6,669 ops/sec ±4.90% (9 runs sampled)
bn.js#pow x 7,335 ops/sec ±3.60% (9 runs sampled)
```

I do see a bigger difference between `mul` & `imul` on its own. That should be documented though if it cannot be improved. One would assume the in-place versions are faster.

indutny:
Yeah, I was even thinking about removing them at all. What's the point of having in-place functions if there are not faster? They are definitely not easier, and not safer to use.

Issue #119, opened 2016-02-03
carry in _iaddn, _isubn


fanatid:
[._iaddn](https://github.com/indutny/bn.js/blob/4ef34bdaae8c73a3b616f4e0498d5b119d664aa1/lib/bn.js#L2138) add `num` to `words[0]`, but `num` can be more than `0x03FFFFFF` (26 bits), as result:
```js
var BN = require('bn.js')
new BN(0).addn(0x90000000).muln(2) // <BN: fffffffffffff20000000>
```
`_isubn` also affected to this bug...

indutny:
It can't really be bigger, this function needs an assert.

PR #122, opened 2016-02-04
bn: fix .iaddn & .isubn for numbers > 0x3ffffff

fanatid:
Details #119 
It can be possible add numbers that more than 0x03ffffff (like in [_initNumber](https://github.com/indutny/bn.js/blob/2fbe7a3683aca9a6a09697a9af2ea7a05d42bf06/lib/bn.js#L112)), but it has overhead. If somebody need add this numbers, he can write `.add(new BN(num))`.

### DIFF
--- a/lib/bn.js
+++ b/lib/bn.js
@@ -1883,6 +1883,7 @@
 
   BN.prototype.imuln = function imuln (num) {
     assert(typeof num === 'number');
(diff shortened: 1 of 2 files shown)


indutny:
Landed in 8a7a5da, please take a look at commit log here too ;) Thank you!

PR #123, opened 2016-02-04
bn: introduce small pow, .pown()

axic:


alidavut:
Any update?

axic:
I think technically works (but would need a rebase and confirming it still works), however perhaps the gain wasn't significant enough at the time?

If there's interest getting this merged, I can rebase it.

alidavut:
I'm new at bn.js and  in my first day I needed a function like this. Current ```pow``` function needs a BN instance which is not usable much for standard js numbers, adds extra code. @indutny ?

PR #124, opened 2016-02-05
iaddn, isubn, imuln (continue)

fanatid:
Sorry, just realized today (after merge :cry:), that we can use that I describe in #122.
Is this better than `assert`? If so, I will fix tests.


indutny:
@fanatid hm... I think it is better to assert instead of doing this polyfill. `n` postfixed operation are supposed to be high-performant, and shouldn't really fallback to slow modes in any case.

PR #126, opened 2016-02-12
benchmark: replace seedable fixture generator

fanatid:
Replace sha256-hash based generator to XorShift128+ from package [xorshift.js](https://www.npmjs.com/package/xorshift.js).
By my measurements xorshift128+ faster than sha256-hash based in ~14x.


dcousens:
Optimizing the benchmarks?

fanatid:
Little optimization for generating fixtures, mostly is just another PRNG.

indutny:
Hm... doesn't sound reasonable to me. Benchmarks take quite a lot of time to run anyway, and this PRNG time is just a minuscule of what is spent on code cycles.

indutny:
Sorry!

dcousens:
Another benefit with this was comparing the same seeds across libraries for easy verification or sharing the inline PRNG that is used here.

indutny:
@dcousens is it just abstract reasoning, or is there any practical examples of this need?

dcousens:
@indutny https://github.com/dominictarr/crypto-bench/pull/6

fanatid:
@dcousens current prng also support seed :)
but xorshift is published prng, whereas current prng is based on sha256 (I needed something for seed, but haven't [xorshift.js](https://www.npmjs.com/package/xorshift.js) in that time)

indutny:
Landed in 6582e29, thank you! @dcousens this was convincing :wink:

PR #127, opened 2016-02-13
bn: avoid returning -0 in .cmp() and .cmpn() (fixes #125)

### LINKED ISSUE #125: .cmp() can return -0
```js
new BN(-42).cmp(new BN(-42)) // -0
```

This line is the cause: https://github.com/indutny/bn.js/blob/master/lib/bn.js#L2768

It is not a major issue per se, as -0 == 0 in Javascript, thus comparing the output with 0 will return true.

axic:


lib/bn.js
```diff
@@ -2750,7 +2750,7 @@
       var w = this.words[0] | 0;
       res = w === num ? 0 : w < num ? -1 : 1;
     }
-    if (this.negative !== 0) {
+    if (this.negative !== 0 && res !== 0) {
       res = -res;
```
indutny: Perhaps it could be just `res = (-res) | 0`?

lib/bn.js
```diff
@@ -2765,7 +2765,7 @@
     if (this.negative === 0 && num.negative !== 0) return 1;
 
     var res = this.ucmp(num);
-    if (this.negative !== 0) return -res;
+    if (this.negative !== 0 && res !== 0) return -res;
```
indutny: Ditto, `(-res) | 0` (not sure if parens are needed, though)

lib/bn.js
axic: Actually that seems to work (at least on my node):

```js
var res = 0;

var a = -res;
console.log(1/a === -Infinity); // true

var b = -res | 0;
console.log(1/b === -Infinity); // false
```

indutny: Of course it does :) This is not related to node.js version.

axic:

indutny:
Yep, thank you! @axic there are no email notification when you push new commits to PR. I didn't see it, sorry.

indutny:
LGTM, landed. Thank you!

Issue #129, opened 2016-02-13
Breaking changes for the next major release


axic:
(Taking the discussion from #112 to here)

Proposed changes:
- [x] Make `.modn()` return a BN (#112)
- [x] Renaming `.strip()` to `._strip()` is a breaking change too (#105)
- [x] Maybe reviewing the constructor method(s) could lead to such a change (see #90 & #91)
- [ ] Rename `.andln()` to `.andn()` (see the README). Perhaps fold the functionality into `toArrayLike` and make `.andn()` safe in terms of working up to 53 bits
- [ ] Maybe at that stage two's complement could be made more internal (part of constructor and toString) (see #73)
- [ ] Split out the extended functions (saving destroyed bits) off `iusrhn()` into a specific shift right version, because it is only used in `MPrime.split()` and is complex.
- [ ] Remove internal functions from the BN context (good candidates are: `_countBits`, `_zeroBits`, etc. Others could be moved too with passing a `self` variable instead of using this, however that might have a speed penalty or increase?)
- [ ] Perhaps rethink naming/functionality of the following bitwise methods: `setn` (should be renamed `isetn` as it is in-place), `testn`, `maskn`, `bincn`

axic:
I also propose to create a `v5.x` branch to merge these changes in.

axic:
Perhaps also #107 could be discussed as part of this.

fanatid:
Any plans regarding https://github.com/indutny/bn.js/issues/35 as part of next major release ?

fanatid:
@indutny just would like to know your opinion about Buffer ([node](http://nodejs.org/api/buffer.html)/[browser](https://github.com/feross/buffer/)) instead Array here

indutny:
@fanatid for what?

fanatid:
out of curiosity...
as I understand you used Array instead Buffer for zero dependencies, why not Uint8Array for example?
if there were no question about dependencies, would you use Buffer instead Array?

indutny:
@fanatid both Buffer and Uint8Array were slower for the purposes of bn.js . I've experimented with it several times, and at each try it was worse.

indutny:
@fanatid additionally, allocation is a huge problem.

fanatid:
oh, okay, so the main reason is performance? How long ago are you tested?
Array allocation work better? Maybe I'm wrong, but as I remember, Buffer allocate more then requested? (or pre-allocated buffers works in such way, at least in node)

[...]

PR #130, opened 2016-02-13
bn: make .modn() return a BN (fixes #112)

### LINKED ISSUE #112: modn API inconsistency
All mod/div methods return a BN instance, except `modn` which returns a JS Number.

I think it would make sense making the current `modn` internal (prefixing it with _) so that `toString` can keep using it the fastest way, and creating a new `modn` which returns a BN instance.

`divmod` already encapsulates a `modn` result as a BN instance so it should use the new method.

Comments?

axic:


calvinmetcalf:
so I am generally against any breaking changes because I have a lot of modules relying on this and on upgrade I have to upgrade all of them or none of them to avoid conflicting versions of bn (and to avoid having excessive stuff included in browserify bundles). so lets think really hard before doing a major version bump and if we decide we want one we should treat it like the last major version for a long time so make sure we miss nothing.

axic:
@calvinmetcalf please see #112 and #129 for more detail

indutny:
Fixed by introducing `modrn` in da8516f.

PR #131, opened 2016-02-19
Use expand() for increasing the words[] array. Speeds up divmod a tiny bit.

axic:
Prepending the words[] array is used in `ishlnsubmul` (used by `divmod`), `inotn`, `setn` and `bincn`. They have various optimal/non-optimal versions.

Simplifies code a lot. It also seems to improve the speed of div and mod by a tiny fraction (on average it is within the error limit).

```
Old:
bn.js#div x 347,770 ops/sec ±1.70% (9 runs sampled)
bn.js#mod x 353,496 ops/sec ±0.77% (8 runs sampled)

New:
bn.js#div x 361,769 ops/sec ±1.12% (9 runs sampled)
bn.js#mod x 362,928 ops/sec ±1.48% (9 runs sampled)
```

### DIFF
--- a/lib/bn.js
+++ b/lib/bn.js
@@ -797,9 +797,7 @@
     var bitsLeft = width % 26;
 
     // Extend the buffer with leading zeroes
-    while (this.length < bytesNeeded) {
-      this.words[this.length++] = 0;
-    }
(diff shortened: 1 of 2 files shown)


indutny:

PR #134, opened 2016-03-24
remove check for function named 'BN'

### LINKED ISSUE #133: BN.isBN() breaks code when minified
In the PR https://github.com/indutny/bn.js/pull/109 it was suggested to include `n.constructor.name === 'BN'` in the `isBN()` check. Unfortunately this breaks behaviour with minified code.

The following sample `elliptic` code runs properly browserified, but fails when minified:
```js
var crypto = require('crypto')
var EC = require('elliptic').ec
var ec = new EC('secp256k1')

var privateKey = crypto.randomBytes(32)
console.log(ec.keyFromPrivate(privateKey).getPublic(true, true))
```

The log output is:
```
Uncaught TypeError: Cannot read property 'fromRed' of null
n.getX @ test.br.min.js:490
n._encode @ test.br.min.js:400
n.encode @ test.br.min.js:401
i.getPublic @ test.br.min.js:588
i.57.crypto @ test.br.min.js:1060
n @ test.br.min.js:6
e @ test.br.min.js:7
(anonymous function) @ test.br.min.js:8
```

The `isBN()` section in the BN constructor should be there to have a quick path, but should we assume the other code path should handle a BN input properly (by toString() and parsing that)?

Had no time to fully track down everything. Perhaps this is more of a bug in `elliptic` where it assumes passing through BN objects will keep the reduction context (which it does if `isBN()` deems it to be an instance, otherwise the reduction context is lost).

calvinmetcalf:
fix #133 and thus crypto-browserify/browserify-sign#27

### DIFF
--- a/lib/bn.js
+++ b/lib/bn.js
@@ -56,7 +56,7 @@
 
   BN.isBN = function isBN (num) {
     return num !== null && typeof num === 'object' &&
-      num.constructor.name === 'BN' && Array.isArray(num.words);
+      num.constructor.name === BN.name && Array.isArray(num.words);
   };
 
   BN.max = function max (left, right) {


jprichardson:
utACK.

calvinmetcalf:
ut?

jprichardson:
> ut?

ut => "untested" :) i.e. the code looks good, but I haven't actually checked out the code and ran tests personally. I think I picked up the term from the Bitcoin repo.

calvinmetcalf:
ah

lib/bn.js
```diff
@@ -56,7 +56,7 @@
 
   BN.isBN = function isBN (num) {
     return num !== null && typeof num === 'object' &&
-      num.constructor.name === 'BN' && Array.isArray(num.words);
+      num.constructor.name === BN.name && Array.isArray(num.words);
```
indutny: You know what I learned recently? This thing does not work in IE: https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Function/name .

Looks like we may need to fallback to using just `Array.isArray()`.

axic: You could still have a constant exported in BN and check against that.

indutny: Like some magic value?

axic: Yes.

Not entirely sure if it would make sense using a function reference, but it might be handy: ```num.isBN === BN.isBN```

This shouldn't work if `num` is an instance of a different BN - do we want that to work or fail?

indutny: If we wanted to check against just one instance of `BN` - we would use `obj instance of BN`, right? :wink:

axic: Well that seems to have good browser support :)

The question is still there: that will still reject BN instances of different versions.  I am pretty sure that in some browserified instances that will happen.  I think it is the desired behaviour, but I don't think many libraries are prepared for this.

axic:
@calvinmetcalf @jprichardson @indutny in the previous in line comment, as a suggestion this came up: `num instance of BN`.

Would that work in all cases? It rejects any instances from a different BN version, which I think is fine, but has its own drawbacks - I assume many projects aren't prepared for that either.

fanatid:
@axis `instanceof` is not good here
https://github.com/bitpay/bitcore-lib/issues/21

axic:
@indutny @calvinmetcalf @jprichardson @fanatid ping. A decision and fix could be useful ;)

jprichardson:
I've reversed my position on this. I think `function.constructor.name` is a good idea to keep. Uglify allows you to not touch this `--keep-fnames` and IE users can always shim. The nice thing is that different `BN.js` dependencies than can be used and we don't get caught with `instanceof` bugs. My two cents.

calvinmetcalf:
@jprichardson yeah but most/all people don't use --keep-fnames so it would just end up broken in reality, so I'd be in favor of this

indutny:
Considering recent stability of `bn.js` versions, I'm inclined to say that we may want to just do `instanceof`, and add an assert like this:

```js
assert(other.name === this.constructor.name && Array.isArray(other.words) && !(other instanceof BN), 'Looks like you have a version mismatch');
```

What do you guys think about this?

calvinmetcalf:
+1 I like it

axic:
@calvinmetcalf @indutny I think this can be closed then?

indutny:
I think yes, thank you!

PR #44, opened 2015-03-20
added fix or pow(base, 0)

wanderer:

PR #135, opened 2016-03-25
Fix incorrect result from redPow with exponent = 0 and modulus = 1

dchest:
In PR #44 fix for the issue with zero exponent missed the case when
modulus is 1: pow(x, y) % 1 will always produce 0, not 1.

### DIFF
--- a/lib/bn.js
+++ b/lib/bn.js
@@ -3280,7 +3280,7 @@
   };
 
   Red.prototype.pow = function pow (a, num) {
-    if (num.isZero()) return new BN(1);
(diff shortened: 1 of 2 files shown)


dchest:
Question: do Red methods support modulus 1 at all?

indutny:
@dchest I believe they don't. Perhaps it is worth documenting it, instead of fixing one occurrence of such error. Thank you!

dchest:
@indutny in this case, maybe add an assert in Red constructor?

indutny:
Sounds like a good idea!

dchest:
Good, here's PR #138

PR #137, opened 2016-03-25
Test fixes

dchest:
* Fixed `assert.throw` tests, which were catching wrong exceptions.
* Removed extra arguments passed to `redSqrt`.

### DIFF
--- a/test/arithmetic-test.js
+++ b/test/arithmetic-test.js
@@ -599,12 +599,12 @@ describe('BN.js/Arithmetic', function () {
     });
     it('should not allow 0 input', function () {
       assert.throws(function () {
-        BN(1).egcd(0);
+        new BN(1).egcd(0);
       });
     });
     it('should not allow negative input', function () {
       assert.throws(function () {
-        BN(1).egcd(-1);
+        new BN(1).egcd(-1);
       });
     });
   });
--- a/test/arithmetic-test.js
+++ b/test/arithmetic-test.js
@@ -81,7 +81,7 @@ describe('BN.js/Arithmetic', function () {
     it('should throw error with num eq 0x4000000', function () {
       assert.throws(function () {
         new BN(0).iaddn(0x4000000);
-      });
+      }, /^Error: Assertion failed$/);
     });
   });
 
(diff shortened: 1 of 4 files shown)


test/constructor-test.js
```diff
@@ -27,7 +27,7 @@ describe('BN.js/Constructor', function () {
       var num = Math.pow(2, 53);
 
       assert.throws(function () {
-        BN(num, 10);
+        num = new BN(num, 10);
```
axic: Is this needed?

test/arithmetic-test.js
```diff
@@ -599,12 +599,12 @@ describe('BN.js/Arithmetic', function () {
     });
     it('should not allow 0 input', function () {
       assert.throws(function () {
-        BN(1).egcd(0);
+        new BN(1).egcd(0);
```
axic: Good catch! Maybe it would be better checking for the actual exception message to fully verify what caused the throw?

fanatid: @dchest please add `/^Error: Assertion failed$/`

test/constructor-test.js
dchest: `num =` ? Yes, otherwise pre-commit check complains about using `new` for side-effects.

fanatid:
It's strange that `semistandard` say nothing about `BN` without `new`.

test/constructor-test.js
fanatid: @dchest you can write `return` instead `num`

axic: And it doesn't complain on the other ones where no assignment is made?

dchest: OK, will do. No, others have other methods in chain, so it doesn't complain.

axic: Oh I see!

fanatid:
@dchest can you also add second argument for other `assert.throws`? :) (my grep expression found 9 matches)

test/utils-test.js
```diff
@@ -202,7 +202,7 @@ describe('BN.js/Utils', function () {
       assert.equal(new BN(0x3fffffe).cmpn(0x3fffffd), 1);
       assert.throws(function () {
         new BN(0x3fffffe).cmpn(0x4000000);
```
fanatid: also need `return`

dchest: It works because there's `cmpn` after `new`, so there's side-effect-only constructor, which `semistandard` reports about.

fanatid: thanks for explanation!

fanatid:
LGTM

indutny:
Landed in c954522, thank you!

PR #140, opened 2016-03-30
add toUnsafeNumber

fanatid:
Why?

If you need number from `bn.js` and you don't care about accuracy.

### DIFF
--- a/README.md
+++ b/README.md
@@ -58,6 +58,7 @@ either `le` (little-endian) or `be` (big-endian).
 * `a.clone()` - clone number
 * `a.toString(base, length)` - convert to base-string and pad with zeroes
 * `a.toNumber()` - convert to Javascript Number (limited to 53 bits)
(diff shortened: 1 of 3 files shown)


indutny:
@fanatid may I ask you for a particular use case example? Where are you going to use it?

fanatid:
I think I created this PR for [fanatid/bencode.js](https://github.com/fanatid/bencode.js)

indutny:
@fanatid mm... and how does it work without proper precision?

fanatid:
[bep_0044](http://bittorrent.org/beps/bep_0044.html) says that number in bencode can be up to MAX_INT64 (63 bits), but JavaScript supports only 53-bits numbers.
but probably I forgot for what I really created this PR... :(

indutny:
Exactly, I guess you need to extract two 32-bit words out of BN to write them in a BEP-way. Sorry, closing PR.

PR #141, opened 2016-03-30
Don't accept invalid characters

dchest:
Contructor from string accepted characters out of base range.
Fix by throwing exception when such character appears in the string.

_PS Spent the whole day trying to figure out why my code didn't work only to discover that I forgot to add 16 as base for constructor (similar to the test in this commit) :sweat_smile:_

### DIFF
--- a/lib/bn.js
+++ b/lib/bn.js
@@ -235,6 +235,7 @@
 
   function parseBase (str, start, end, mul) {
     var r = 0;
+    var b = 0;
     var len = Math.min(str.length, end);
     for (var i = start; i < len; i++) {
       var c = str.charCodeAt(i) - 48;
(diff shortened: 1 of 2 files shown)


fanatid:
Related: https://github.com/indutny/bn.js/issues/90 https://github.com/indutny/bn.js/pull/91

dchest:
@fanatid this one will just make sure that decoded number is less than base, but won't reject other characters, e.g. I think `:` and `;`, etc. (they are between numbers and letters in ASCII) will be accepted for base 16.

fanatid:
semver-major btw

indutny:
Landed in 9827640, thank you!

PR #152, opened 2016-11-07
Add section files to package.json

fanatid:
White list is better than `.npmignore`
related issue: #150


dcousens:
Rebase?

indutny:
Landed similar change in b267676. Thank you!

PR #162, opened 2017-05-16
doc - add note clarifying n-suffix and decimals

kumavis:


indutny:
Landed in 7d14aaf, thank you!

PR #164, opened 2017-05-23
toJSON to default to interoperable hex (length % 2)

dcousens:
Thoughts?
This catches me out quite often and I always end up having to then roll in `.toString(16, 2)` rather than just letting `toJSON` work it.

### DIFF
--- a/lib/bn.js
+++ b/lib/bn.js
@@ -510,7 +510,7 @@
   };
 
   BN.prototype.toJSON = function toJSON () {
-    return this.toString(16);
+    return this.toString(16, 2);
   };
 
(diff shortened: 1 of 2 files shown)


axic:
Why is this called `toJSON`? JSON doesn't even support hex numbers, so it should be base-10 if we call it `toJSON`?

dcousens:
Great point.   I assume it was so eventually a `fromJSON` would import hex only?

dcousens:
@indutny thoughts?

indutny:
This is definitely a breaking change, but it is time to do some. Let's land this and few other major PRs.

indutny:
Landed in a46d2e2, thank you!

PR #166, opened 2017-06-20
redPow should alway return a red (fixes #165)

### LINKED ISSUE #165: redPow(0) returns a regular BN instead of a red

fp-x:


indutny:
Thank you, landed!

PR #167, opened 2017-06-24
bn: improve allocation speed for buffers. avoid deprecation warning.

chjj:
Node.js 8 calls out to `Buffer.alloc` when invoking `new Buffer()`. This ensures `Buffer.allocUnsafe` is called for some extra speed.


indutny:
Landed in 9f7d21f, thank you!

PR #168, opened 2017-07-14
Properly exclude "buffer" from Webpack builds

swansontec:
The technique of `require('buf' + 'fer')` doesn't actually fool Webpack.

The proper way to exclued a module is with the following `package.json` field:

https://github.com/defunctzombie/package-browser-field-spec


indutny:
Landed, thank you!

PR #174, opened 2018-02-02
Add more tests to divRound with negative numbers

axic:


axic:
These tests actually fail, because contrary to the name, the code seems to be rounding to ceiling.

dcousens:
@axic ceiling? or towards `0`?

axic:
Towards zero.

dcousens:
@axic anything left to do here?

axic:
I don't particularly have a lot of time, but I guess this PR could be changed to add tests for the current behaviour and update documentation to reflect that.

A separate PR could change behaviour as that is breaking - well, it could be just a bug too, if the document is to be the canon. Any opinions @indutny ?

Issue #177, opened 2018-02-15
In-place operation in a reduction context is not reduced.


mahrud:
Seems like in-place operations are not reducing the result before changing the host object. Example:
```
var BN=require('bn.js')
red = new BN.red(new BN(13))
a = new BN(2).toRed(red)
b = new BN(7).toRed(red)
x = a.redIMul(b)

red.m // <BN: d>
x // <BN-R: 1>
x.cmp(red.m) // correct
a // <BN-R: e>
a.cmp(red.m) // incorrect
```
Note that `a` is still in a reduction context, but its value is larger than the modulus!

fanatid:
This happens because `Red#imod` is not actually `imod`... :( https://github.com/indutny/bn.js/blob/98ac84e81dc02d178168460e5889c80840853a44/lib/bn.js#L3172

dcousens:
Closing in favour of #178

PR #178, opened 2018-02-16
bn: fix Red#imod

fanatid:
Issue #177

### DIFF
--- a/lib/bn.js
+++ b/lib/bn.js
@@ -327,6 +327,13 @@
     dest.red = this.red;
   };
 
+  BN.prototype.move = function move (dest) {
+    dest.words = this.words;
+    dest.length = this.length;
+    dest.negative = this.negative;
+    dest.red = this.red;
+  };
+
   BN.prototype.clone = function clone () {
     var r = new BN(null);
     this.copy(r);
@@ -3170,8 +3177,7 @@
   Red.prototype.imod = function imod (a) {
     if (this.prime) return this.prime.ireduce(a)._forceRed(this);
 
-    var mod = a.umod(this.m)._forceRed(this);
-    if (mod !== a) mod.copy(a);
+    a.umod(this.m)._forceRed(this).move(a);
     return a;
   };
 
--- a/lib/bn.js
+++ b/lib/bn.js
@@ -3169,7 +3169,10 @@
 
   Red.prototype.imod = function imod (a) {
     if (this.prime) return this.prime.ireduce(a)._forceRed(this);
-    return a.umod(this.m)._forceRed(this);
+
+    var mod = a.umod(this.m)._forceRed(this);
+    if (mod !== a) mod.copy(a);
+    return a;
   };
 
   Red.prototype.neg = function neg (a) {
--- a/lib/bn.js
+++ b/lib/bn.js
@@ -327,7 +327,7 @@
     dest.red = this.red;
   };
 
-  BN.prototype.move = function move (dest) {
+  BN.prototype._move = function _move (dest) {
     dest.words = this.words;
     dest.length = this.length;
     dest.negative = this.negative;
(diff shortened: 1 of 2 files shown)


lib/bn.js
```diff
@@ -3169,7 +3169,10 @@
 
   Red.prototype.imod = function imod (a) {
     if (this.prime) return this.prime.ireduce(a)._forceRed(this);
-    return a.umod(this.m)._forceRed(this);
+
+    var mod = a.umod(this.m)._forceRed(this);
+    if (mod !== a) mod.copy(a);
```
indutny: This might be not so good performance-wise. Any ideas?

indutny: Should we introduce `iumod` instead?

fanatid: Agree that this can introduce performance issues... I thought about new method `.move` instead `.copy`.

> Should we introduce `iumod` instead?

do you mean `BN#iumod`?

indutny: `.move` might actually work great here!

fanatid: Do you want it as separate method or only as few lines in `Red#imod`? What about `BN#iumod`? It's doesn't cost it?

dcousens: @fanatid do you mean new method for `.move`? If so,  I think yes

fanatid: `BN#move` added in https://github.com/indutny/bn.js/pull/178/commits/fa40b8c7cf349eebffb825ad67cc554462cf1f33

lib/bn.js
```diff
@@ -3169,7 +3176,10 @@
 
   Red.prototype.imod = function imod (a) {
     if (this.prime) return this.prime.ireduce(a)._forceRed(this);
-    return a.umod(this.m)._forceRed(this);
+
+    var mod = a.umod(this.m)._forceRed(this);
+    if (mod !== a) mod.move(a);
```
dcousens: is it even worth checking if they are different? (performance diff?)

test/red-test.js
```diff
@@ -262,4 +262,13 @@ describe('BN.js/Reduction context', function () {
     red.prime.split(input, output);
     assert.equal(input.cmp(output), 0);
   });
+
+  it('imod should change host object', function () {
```
dcousens: maybe indicate `imod` (via `redIMul`)

lib/bn.js
fanatid: In one case `mod` === `a`. Not sure does it cost check it or not... https://github.com/indutny/bn.js/blob/98ac84e81dc02d178168460e5889c80840853a44/lib/bn.js#L2413

dcousens: right,  if `mod === a`, no need to move,  but,  as `move` guarantees that `a == mod`,  why bother checking at all?
It looks like an optimization,  but,  IMHO serves to confuse.

fanatid: Changed to `a.umod(this.m)._forceRed(this).move(a);`

lib/bn.js
```diff
@@ -327,6 +327,13 @@
     dest.red = this.red;
   };
 
+  BN.prototype.move = function move (dest) {
```
indutny: Oh, shouldn't it be private? It looks like this could lead to funny state issues if used as a part of API.

fanatid: I'll not mind of making it private. Do you want see it as separate function or as "inline" function?

indutny: IMO, it makes sense to have it on a prototype. Just prefix it with `_`. @dcousens what do you think?

dcousens: @indutny I think the public interface semantics of `move` can be assumed that the originating object is reset or garbage after use.

dcousens: @fanatid prefix it with `_`?

fanatid: @dcousens added prefix `_`, let me know if I should squash commits

PR #179, opened 2018-03-05
Adds typescript definition file.

MicahZoltu:
This isn't complete, but it is better than nothing and provides future users a place to easily extend the definitions.

### DIFF
--- a/lib/bn.d.ts
+++ b/lib/bn.d.ts
@@ -1,0 +1,17 @@
+declare module 'bn.js' {
+    class BN {
+        constructor(value: string | number, radix?: number);
+        toString(radix: number): string;
+        toNumber(): number;
+        add(other: BN): BN;
+        sub(other: BN): BN;
+        mul(other: BN): BN;
+        div(other: BN): BN;
+        pow(other: BN): BN;
+
+        static min(... args: Array<number|string|BN>): BN;
+        static max(... args: Array<number|string|BN>): BN;
+    }
+
+    export = BN;
+}
--- a/package.json
+++ b/package.json
@@ -3,6 +3,7 @@
   "version": "4.11.8",
   "description": "Big number implementation in pure javascript",
   "main": "lib/bn.js",
+  "types": "lib/bn.d.ts",
   "scripts": {
     "lint": "semistandard",
     "unit": "mocha --reporter=spec test/*-test.js",


vsund:
Hey @MicahZoltu, I found a [repository with type declarations](https://github.com/machinomy/types-bn) for `bn.js` while searching for types.

You can probably can add the missing types from there :)

MicahZoltu:
I think this repository/project may be dead, pulse indicates no one has done anything for over a month, and this PR has been sitting open for over a week with nothing from the repo owner.  My recommendation would be for someone to submit a PR to DefinitelyTyped with the definitions (probably a combination of what is in that other repo and what is here).

I don't have a pressing need at the moment, so I probably won't do it myself, but consider everything I have here to be CC0 or WTFPL or Unlicense (i.e., feel free to crib it without attribution).

dcousens:
@MicahZoltu the project is certainly not dead.  A lack of activity on this PR probably indicates a simple lack of interest.

MicahZoltu:
Glad to hear the project isn't dead.  😄 To be clear, my assumptions of life were based on GitHub's project pulse, which shows this project hasn't seen any changes for over a month, not based on lack of response to this issue.

Should I assume this to mean that you aren't interested in getting type definitions included and those of us who want type definitions should go another route (like DefinitelyTyped)?

dcousens:
@MicahZoltu the perpetual issue is maintenance,  I personally don't use typescript,  and I don't know about the other contributors,  but a lack of activity on this PR argues for the same.

`DefinitelyTyped` is probably the least-terrible approach,  even though it is external.
Unless @indutny is happy to merge this,  I'd suggest you merge there,  and then let us put a link in the README here.

MicahZoltu:
Generally speaking, if it is possible to get the TypeScript definitions into the project's NPM package that is superior for a number of reasons including versioning, simplicity and maintainability.  Definitely Typed is used when, for whatever reason, it is not possible to get the definitions into the project's NPM package directly.

Even if the authors of this package don't use TypeScript or the definitions, having the definitions here is superior (IMO) to putting them in Definitely Typed, which is why I opened this PR first.  That being said, if the maintainers of this project are uncomfortable with the addition of the TypeScript definitions (which it sounds like is the case) then Definitely Typed is a reasonable fallback strategy.

indutny:
@MicahZoltu I've been playing a lot with TS recently and am very happy with type checks there. However, the issue here is indeed maintenance of the mentioned type definitions.

I've a suggestion: we've add a `ts-node` as a dev dependency and introduce a test that would cover all API using the types. This test will be used by maintainers for determining whether the release breaks the types or not.

If it breaks - an issue will be opened and the release will be delayed for maximum 30 days. If no one will come to help us - we remove the type file.

@dcousens does it sound reasonable?

dcousens:
@indutny sounds reasonable.

MicahZoltu:
Minor correction for clarity, the test itself likely wouldn't break, but compilation of the test would.

If the maintainers aren't actually using TS themselves, then in my experience releasing with broken type definitions _may_ be a better option.  When projects do this, what happens is some TypeScript user pulls down the latest version of the library, finds the definitions don't align with reality, then they submit a PR to update the definitions.  These people likely aren't monitoring this repository's GitHub issues so they are unlikely to become aware of the broken definitions in a pre-release build.

That being said, a change that breaks the definitions should be a SemVer major upgrade as it is backward incompatible.  Even if the maintainers aren't keeping the definitions up-to-date, having some kind of automated test/compilation that fails when they break can be a good indicator to the maintainer that they should do a SemVer major update.

dcousens:
> finds the definitions don't align with reality

How does this occur?
Blind faith in the underlying types can be dangerous when this library is used for cryptography.

indutny:
@dcousens it'd throw on compilation. So, I guess, I agree with @MicahZoltu that for non-breaking API changes we may ship with broken typedefs.

MicahZoltu:
It depends on the change.  If some function is changed to become more permissive in the set of inputs it accepts, then the old types will still be fine as they are more restrictive than reality.  If a function is changed such that it accepts fewer parameters, different parameters, or returns a different type then the engineer would either get a compilation error or a runtime error depending on whether the developer is following the types (runtime error) or they are following the documentation (compile time error).

When authoring automated tests, for return types you would want to have the test assert that the returned value has a runtime type that aligns with the type definition.  For parameters, you would want to supply parameters that align with the type definition and assert that the function runs as expected.  These tests would need to be written in TypeScript so compilation of them will fail if the type definitions don't line up with the test and assertions will fail if runtime types don't align with the expected runtime type (which should match the TS definition).

The problem with all of this (automated testing of definitions) of course is that it doesn't really test what you want it to test, instead it tests that the author of the test asserted the right types, not that the asserted type aligns with the definition.  The solution to this problem is to actually write the library in TypeScript.  😄 Since that is a significantly larger undertaking compared to authoring a definition file, people usually just don't test definition files for JS libraries.

vsund:
Glad to have some activity in this :)

I'd like to throw in [`dtslint`](https://github.com/Microsoft/dtslint) in here. Although I never used it, it seems to be what we need for checking/linting type declarations.

MicahZoltu:
I believe dtslist is only a linter (style) not a static analyzer (correctness)?  I'm not against it, but it feels like overkill for a single small `.d.ts` file.

MicahZoltu:
Superceded by #180

PR #180, opened 2018-03-21
Add TypeScript declarations

vsund:
This PR adds type declaration based on https://github.com/indutny/bn.js/pull/179 and [`types-bn`](https://github.com/machinomy/types-bn), extends these both a bit and adds a test frame as discussed in #179.

---
Probably supersedes https://github.com/indutny/bn.js/pull/179 and [`types-bn`](https://github.com/machinomy/types-bn).

cc @indutny @dcousens @MicahZoltu

### DIFF
--- a/lib/bn.d.ts
+++ b/lib/bn.d.ts
@@ -1,0 +1,66 @@
+/*
+ * These types are based on [1] from @ukstv and [2] from @MicahZoltu.
+ * If anything is wrong with these types, feel free to complain to @vsund.
+ * 
+ * [1] https://github.com/machinomy/types-bn
+ * [2] https://github.com/indutny/bn.js/pull/179
+ */
+
+
+export declare type Endianness = 'le' | 'be'
+
+export declare class BN {
+  static isBN(b: any): boolean;
+  static min(left: BN, right: BN): BN;
+  static max(left: BN, right: BN): BN;
+
+  constructor(number: number | string | number[] | Buffer, base?: number, endian?: Endianness);
+  clone(): BN;
+  copy(dest: BN): void;
+  inspect(): string;
+  toString(base?: number | 'hex', length?: number): string;
+  toNumber(): number;
+  toJSON(): string;
+  toArray(endian?: Endianness, length?: number): number[];
+  toBuffer(endian?: Endianness, length?: number): Buffer;
+  toBitArray(num: BN): { [bit: number]: number };
+  bitLength(): number;
+  zeroBits(): number;
+  byteLength(): number;
+  isNeg(): boolean;
+  isEven(): boolean;
+  isOdd(): boolean;
+  isZero(): boolean;
+  cmp(b: BN): number;
+  lt(b: BN): boolean;
+  lte(b: BN): boolean;
+  gt(b: BN): boolean;
+  gte(b: BN): boolean;
+  eq(b: BN): boolean;
+
+  neg(): BN;
+  abs(): BN;
+  add(b: BN): BN;
+  sub(b: BN): BN;
+  mul(b: BN): BN;
+  sqr(): BN;
+  pow(b: BN): BN;
+  div(b: BN): BN;
+  mod(b: BN): BN;
+  divRound(b: BN): BN;
+
+  or(b: BN): BN;
+  and(b: BN): BN;
+  xor(b: BN): BN;
+  setn(b: number): BN;
+  shln(b: number): BN;
+  shrn(b: number): BN;
+  testn(b: number): boolean;
+  maskn(b: number): BN;
+  bincn(b: number): BN;
+  notn(w: number): BN;
+
+  gcd(b: BN): BN;
+  egcd(b: BN): { a: BN, b: BN, gcd: BN };
+  invm(b: BN): BN;
+}
--- a/lib/bn.d.ts
+++ b/lib/bn.d.ts
@@ -23,7 +23,6 @@ export declare class BN {
   toJSON(): string;
   toArray(endian?: Endianness, length?: number): number[];
   toBuffer(endian?: Endianness, length?: number): Buffer;
-  toBitArray(num: BN): { [bit: number]: number };
   bitLength(): number;
   zeroBits(): number;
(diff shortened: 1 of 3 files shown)


vsund:
Also as brought up in #179, I just added a `npm run test-ts-declarations` and not added it to any build task or similar. Maintainers should test this manually.

Let me know if I should change this.

vsund:
Should be complete now. The code is extremely repetitive and probably could be refactored to test a set of methods with the same number of parameters (all unary number, all binary BN, etc).

Is this what you meant @indutny?

vsund:
Just played a bit around with refactoring.

```ts
const bn1 = new BN(1);

console.log('Testing now methods without parameters that return a number');
const methods: number[] = [
    bn1.bitLength(),
    bn1.zeroBits(),
    bn1.byteLength()
];
for(const method of methods) {
    if(typeof method !== 'number') {
        throw new Error('');
    }
}
```
was the closest I could get. Not sure whether this would be better.

Also all the type checks are probably not needed, the code won't compile if there's errors (as @MicahZoltu said somewhere in https://github.com/indutny/bn.js/pull/179).

ovr:
ping @indutny

lib/bn.d.ts
```diff
@@ -0,0 +1,65 @@
...
+
+
+export declare type Endianness = 'le' | 'be'
+
+export declare class BN {
+  static isBN(b: any): boolean;
+  static min(left: BN, right: BN): BN;
+  static max(left: BN, right: BN): BN;
+
+  constructor(number: number | string | number[] | Buffer, base?: number | 'hex', endian?: Endianness);
```
webmaster128: Uint8Array is also a valid type for number, right?

webmaster128:
bn.js recently got into DefinitlyTyped: https://github.com/DefinitelyTyped/DefinitelyTyped/pull/26120

The definition file there is a starting point, but not at all complete. There are also some problematic definitions like `gt(b: any): boolean;`, which is a trap for users putting in `number`s.

It would be great to see this work be merged into DefinitlyTyped to make it easier for users to find the one well maintained version.

webmaster128:
In order to use the definitions together with existing `import BN = require("bn.js");` import statements, I had to wrap everything as follows

```
declare module 'bn.js' {
  type Endianness = 'le' | 'be'

  class BN {
    static isBN(b: any): boolean;
    static min(left: BN, right: BN): BN;
    static max(left: BN, right: BN): BN;

    constructor(number: number | string | number[] | Buffer | Uint8Array, base?: number | 'hex', endian?: Endianness);

    // all the stuff …

  }

  export = BN;
}
```

For some reason I don't understand, [@types/bn.js](https://github.com/0xProject/DefinitelyTyped/blob/master/types/bn.js/index.d.ts) does not require the `declare module 'bn.js'` part. In any case, the commonjs version of bn.js exports a single onject.

lib/bn.d.ts
vsund: Not tried this out, but from looking at the code I'd say yes :)

webmaster128: I am using Uint8Array inputs in a project with a bunch of tests and do not have any issues with it.

vsund:
Hey @webmaster128,

> It would be great to see this work be merged into DefinitlyTyped to make it easier for users to find the one well maintained version.

I'd rather like this merged here, since it's closer to the actual code and @indutny showed [interest in merging this](https://github.com/indutny/bn.js/pull/179#issuecomment-373894379). Technically this is ready for merge.

On the import statement: I haven't used other import styles than the default `import * as foo from 'foo';`, I remember that I had problem with this import style and one of the type declarations I based this on. Unfortunately I currently don't have much time. Would you like to pick this up @webmaster128?

webmaster128:
okay, the `declare module 'bn.js'` wrapper is a minor issue. This is just a matter of lookup and when the types are bundled with the npm package, we don't have the issue anymore.

The interesting part now is that this library behaves differently in different module systems because of this code:

```javascript
  if (typeof module === 'object') {
    // in a commonjs context
    // the entire module is the BN class
    module.exports = BN;
  } else {
    // other context (browser?)
    // `exports` is the `this` value of the library file, i.e. the global scope
    exports.BN = BN;
  }
```

Since the library contains no ECMAScript 2015 export declarations, I think we need TypeDefinitions for the commonjs case. This is described here

> Both CommonJS and AMD generally have the concept of an exports object which contains all exports from a module.
>
> They also support replacing the exports object with a custom single object. Default exports are meant to act as a replacement for this behavior; however, the two are incompatible. TypeScript supports export = to model the traditional CommonJS and AMD workflow.
> 
> The export = syntax specifies a single object that is exported from the module. This can be a class, interface, namespace, function, or enum.
>
> When exporting a module using export =, TypeScript-specific import module = require("module") must be used to import the module.

https://www.typescriptlang.org/docs/handbook/modules.html

The import method `import * as foo from 'foo';` should not work for this library because

> An ES6 module namespace object cannot be invoked as a function or with new.

https://stackoverflow.com/a/39415662

In our case, the module namespace object is `BN`. Thus TypeScript's `import module = require("module")` must be used instead of ES6's `import * as foo from 'foo';`.

So I think we need

```
type Endianness = 'le' | 'be'

class BN {
  static isBN(b: any): boolean;
  static min(left: BN, right: BN): BN;
  static max(left: BN, right: BN): BN;

  constructor(number: number | string | number[] | Buffer | Uint8Array, base?: number | 'hex', endian?: Endianness);

  // all the stuff …

}

export = BN;
```

webmaster128:
You export style actually works because of this little trick in the library (last line):

```javascript
  if (typeof module === 'object') {
    module.exports = BN;
  } else {
    exports.BN = BN;
  }

  BN.BN = BN;
```

However, this differs from the documented import path:

```javascript
// documented
const BN = require('bn.js');

// what these type definitions do
const BN = require('bn.js').BN;
```

vsund:
Will going to close this in favor of https://github.com/indutny/bn.js/pull/190. Thanks for picking this up @webmaster128!

PR #187, opened 2018-06-07
Trailing slash required for non-node buffer

earonesty:
This breaks on IOS and some browsers.


earonesty:
Looks like the real issue is that built-in Buffer and the module Buffer aren't identical.   I think I figured out the issue.

PR #188, opened 2018-06-08
istanbul gives you wings

knoxcard:


fanatid:
Istanbul is not used currently in project. This dev dependency will be removed while commands for coverage not in package.json:scripts

PR #190, opened 2018-06-22
Add TypeScript definition file

webmaster128:
This is a version of #180 with the following additions

* Make library a single constructor module (like [@types/bn.js](https://github.com/0xProject/DefinitelyTyped/blob/master/types/bn.js/index.d.ts) and the documented import style `const BN = require('bn.js');`
* Add Uint8Array support in constructor and toArrayLike
* Explicitly compile and run tests instead of using ts-node
* Add missing type dependency for nodejs's Buffer

as @vsund suggested, I'd pick up his work and make this ready for merging


lib/bn.d.ts
```diff
@@ -0,0 +1,71 @@
...
+ * 
+ * [1] https://github.com/machinomy/types-bn
+ * [2] https://github.com/indutny/bn.js/pull/179
+ */
+
+
+declare type Endianness = 'le' | 'be'
+
+declare class BN {
+  static isBN(b: any): boolean;
```
jacogr: Would prefer 

```
static isBN(b: any): b is BN;
```

webmaster128: thx, updated

webmaster128:
@jacogr Thanks for feedback! I extended the types by what I found in the README. Your examples should be coveretd now. I'd appreciate another review. If something is missing, please let me know explicitly.

Regarding reductions contexts: I never work with those and would be happy if someone can step in to provide commits I can cherry-pick or builds on top of this PR.

webmaster128:
The types at DefinitelyTyped got improved a lot recently. I was able to use them for a big project with minor code changes (https://github.com/iov-one/iov-core/pull/1397). 

Closing this in favour of DT. Please submit the missing interfaces there:
- https://github.com/DefinitelyTyped/DefinitelyTyped/tree/master/types/bn.js
- https://npmjs.com/package/@types/bn.js

Issue #215, opened 2019-02-25
About signed zero


usd-yamazaki:
Is this correct?

```
var zero = new BN(0);
zero.toString();
> '0'
zero.subn(1).addn(1).toString();
> '-0'
```

I expect the following behavior.

```
var zero = new BN(0);
zero.toString();
> '0'
zero.subn(1).addn(1).toString();
> '0'
```

fanatid:
fixed in #216

PR #216, opened 2019-02-25
fix iaddn sign issue

fanatid:
Issue #215 

When we sub number from BN instance and this number equal to number in
BN instance we receive zero with negative sign.


axic:
You may want to add a test case.

fanatid:
@axic thanks for remainder, fixed :)

PR #220, opened 2019-07-07
Update dev dependencies

fanatid:
Not a big deal, but wanted keep at actual stage for development.

- `istanbul` removed, if coverage will be required will need add scripts and add `nyc` as dependency
- `semistandard` updated and new error messages fixes, probably new version will out soon (new `standard` version was released 2 days ago)
- `mocha` updated, nothing changed

Added `yarn.lock` for locking dev dependencies, [travis](https://travis-ci.org/) now install dependencies with `yarn`.

Node.js versions in travis config updated, it's not possible keep 7,9 because one of dependencies of dependencies have restrictions in `engines` field, details: https://travis-ci.org/fanatid/bn.js/builds/555443389
Version 6 is not possible, because `assert` package at that time did not had `strict`version (`require('assert').strict`). I think it's possible to solve by using `strictEqual` and other, but is this matter run tests on version 6?

Going squash and merge after 24h if there will no NACKs.


fanatid:
Last commit need to be fixed a little, fix in `eslint-plugin-es5`: https://github.com/nkt/eslint-plugin-es5/pull/34

Issue #251, opened 2020-08-13
Handle defined but not implemented Symbol.for


maxfortun:
`Symbol.for` is defined but not implemented in NGINX njs plugin and throws exception.

> err InternalError: not implemented
>     at Symbol.for (native)
> 

http://hg.nginx.org/njs/file/default/src/njs_symbol.c#l174

Suggestion for https://github.com/indutny/bn.js/blob/5707aed66c02ae10f8c7d59f5244fcce1c24cdb6/lib/bn.js#L373

Wrap in try/catch block and revert to `BN.prototype.inspect = inspect;` on error.

PR #252, opened 2020-08-13
Add support for defined but not implemented Symbol.for

maxfortun:
`Symbol.for` is defined but not implemented in NGINX njs plugin and throws exception.

> err InternalError: not implemented
>     at Symbol.for (native)
> 

http://hg.nginx.org/njs/file/default/src/njs_symbol.c#l174

Suggestion for https://github.com/indutny/bn.js/blob/5707aed66c02ae10f8c7d59f5244fcce1c24cdb6/lib/bn.js#L373

Wrap in try/catch block and revert to `BN.prototype.inspect = inspect;` on error.


maxfortun:
#251

fanatid:
LGTM, thank you!

fanatid:
Released as `v5.1.3`.

PR #260, opened 2020-12-14
fix: Buffer not using global in browser, closes #259

### LINKED ISSUE #259: Buffer variable overwrites global Buffer
Edit: Duplicate of https://github.com/indutny/bn.js/issues/255

------------------------

BN throws buffer undefined assertion errors in some browser-like environments, despite having `Buffer` defined as a global. I'm using bn in an electron app with strict security settings: `nodeIntegration: false`, `contextIsolation: true`, `enabledRemoteModule: false` etc.

With these settings `require` is disabled in the renderer thread, where bn.js is invoked. Because of this, `Buffer` needs to be polyfilled as a global.

The following code declares a variable that redeclares the globally defined `Buffer` type to `undefined` in the local scope. Then silently fails when it tries to tries to require in the try block.

https://github.com/indutny/bn.js/blob/6ef485dc30b002ec8100602998763fec8a1a3897/lib/bn.js#L51-L55

Not sure on the most appropriate fix, but something like this should work around the issue.

```
Buffer = window.Buffer || global.Buffer || require('buffer').Buffer;
```

kyranjamie:
Closes issue #259

Redeclaring Buffer prohibits using a globally-defined polyfill.

### DIFF
--- a/lib/bn.js
+++ b/lib/bn.js
@@ -50,7 +50,11 @@
 
   var Buffer;
   try {
-    Buffer = require('buffer').Buffer;
+    if (typeof window !== 'undefined' && typeof window.Buffer !== 'undefined') {
+      Buffer = window.Buffer;
+    } else {
+      Buffer = require('buffer').Buffer;
+    }
   } catch (e) {
   }
 


fanatid:
Thanks. Is it possible to test the new code somehow?

fanatid:
ping @kyranjamie

kyranjamie:
I had a look, but wasn't figure out how to test this. There are no tests are in place at present.

Also ran into issues coercing a mock `window` object, where linter rejects with ` Read-only global 'window' should not be modified.`. Found no docs as to how to disable this check.

fanatid:
Thank you.

Issue #306, opened 2023-12-22
BN.muln with number bigger than 0x4000000


youPickItUp:
BN.muln throws when called with number bigger than 0x4000000

I haven't came across this information until it has blown up on a website :cry: 

I wanted to make sure it's true and ask for some warning / info in the README / docs.
I can do the PR just let me know if that's correct.

fanatid:
This is expected, internally BN operates 26 bits numbers, so you need to use `mul`
https://github.com/indutny/bn.js/blob/5df40f81ea8afb835b909bb7c21e0833cdeb6a30/lib/bn.js#L1993

PR #307, opened 2023-12-27
Document the max plain JS number argument value

youPickItUp:
As mentioned here https://github.com/indutny/bn.js/issues/306 max plain JS number argument value seems to undocumented. I've added info in README.md.


fanatid:
Thank you!

PR #312, opened 2025-04-25
imuln/muln incorrectly handle multiplication by zero

CanardMandarin:
Hello,

I have encountered an issue with the `imuln` function.
Multiplying a BN by zero using `imuln/muln(0)` does not reduce the length.

Steps to reproduce:

```js
> var zero = new BN(0);
> new BN(1111111111).muln(0).eq(zero)
false
```

### DIFF
--- a/lib/bn.js
+++ b/lib/bn.js
@@ -2008,6 +2008,7 @@
       this.words[i] = carry;
       this.length++;
     }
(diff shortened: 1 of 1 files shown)


fanatid:
Thank you! Published as 5.2.2 / 4.12.2

PR #321, opened 2026-07-08
fix: normalize sign when imuln multiplies by zero

spokodev:
Multiplying a negative number by the plain number 0 zeroed the words but left the negative flag set, producing a negative zero. That value reported `isNeg()` as true, was not `eq()` to zero, compared as less than zero, and printed as `-0`. `new BN(-5).mul(new BN(0))` correctly gives `0`; only the `imuln`/`muln` path diverged.

This clears the sign along with the length when the factor is zero so the result is a plain zero. The existing zero-multiplication test only checked `isZero()`, which is true even for a negative zero, so the added test asserts the sign directly.

### DIFF
--- a/lib/bn.js
+++ b/lib/bn.js
@@ -2014,7 +2014,10 @@
       this.words[i] = carry;
       this.length++;
     }
-    this.length = num === 0 ? 1 : this.length;
+    if (num === 0) {
+      this.length = 1;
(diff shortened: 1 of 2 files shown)


fanatid:
Thank you! Both #320 and #321 published in v4.12.5 and v5.2.5

